PR #2, opened 2020-03-13
Rita: added feature test: view peeps

lxierita:


othetgyi:
Hi Rita,

Thanks for letting me a take a look at your code. It's very clean and your controller is very streamlined and doesn't have any unnecessary code. Your model also just has just enough code to do what you want it to. I really loved all the styling on your posts page! You've also nicely separated out all your feature tests. The methods all work, so nice work!

Olivia

spec/setup_test_database.rb
```diff
@@ -7,5 +7,5 @@ def setup_test_database
 
 def add_row_to_test_database
   connection = PG.connect(dbname: 'chitter_test')
-  connection.exec("INSERT INTO peeps (message) values ('This is a peep!');")
+  connection.exec("INSERT INTO peeps (message, date) values ('This is a peep!', '1997');")
```
lindseydew: Nice! Did you make a database migration to change the structure of the table? It's good practise to include that script in the migrations folder so that another dev can reproduce the database.

spec/test_helper.rb
```diff
@@ -0,0 +1,5 @@
+def create_new_post(message)
+	visit '/posts/new'
+	fill_in('message', with: message)
```
lindseydew: Excellent abstraction

views/index.erb
```diff
@@ -0,0 +1,55 @@
...
+		}
+
+	</style>
+</head>
+<body>
+	<h1>Chitter</h1>
+	<% if @peeps.length != 0 %>
+	<ul>
+		<% for @peep in @peeps %>
+			<li>
```
lindseydew: Awesome! 💥

lib/peep.rb
```diff
@@ -0,0 +1,39 @@
...
+		@message = message
+		@date = date
+	end
+	def self.all 
+		if (ENV["ENVIRONMENT"] == 'test')
+			connection = PG.connect(dbname: 'chitter_test')
+		else 
+			connection = PG.connect(dbname: 'chitter')
+		end
+		result = connection.exec("SELECT * FROM peeps ORDER BY id DESC;")
```
lindseydew: Nice! I can see that this would work. Since we're ordering by time it would be more explicit for the query to order by the date field. This could work by making the field type in postgres a timestamp rather than a string.

lib/peep.rb
```diff
@@ -0,0 +1,39 @@
...
+		Peep.new(result[0]['message'])
+	end
+
+	def self.search(keywords)
+		if (ENV["ENVIRONMENT"] == 'test')
+			connection = PG.connect(dbname: 'chitter_test')
+		else 
+			connection = PG.connect(dbname: 'chitter')
+		end
+		result = connection.exec("SELECT * FROM peeps WHERE message LIKE '%#{keywords}%'ORDER BY id DESC;")
```
lindseydew: Lovely keyword search!

app.rb
```diff
@@ -1,9 +1,36 @@
 require 'sinatra/base'
+require './lib/peep'
 
 class Chitter < Sinatra::Base
+	get '/' do
+		'haha gotcha, try type in "/posts" up there&#9757;&#9757;&#9757;&#9757;'
```
lindseydew: lol

app.rb
```diff
@@ -1,9 +1,36 @@
...
+  	posted_date = Time.now.strftime("%d/%m/%Y %H:%M")
+  	Peep.create(message: params[:message], date: posted_date)
+  	redirect '/posts'
+  end
+
+  get '/posts/search' do 
+  	erb :"posts/search"
+  end
+  post '/posts/search' do 
+  	@peeps = Peep.search(params[:keywords])
```
lindseydew: I can see why you've chosen a post here because the browser is sending a parameter to the server. In this case it would be more conventional to use a get request because this method is reading data from the server and doesn't add any new data to the database. In general POST requests are used when new data is added to the server and GET requests are used when data is read from the server. 

But this will work and looks great 👍

PR #9, opened 2020-03-16
Code review

othetgyi:


app.rb
```diff
@@ -1,8 +1,21 @@
...
+  get '/peeps' do
+    #get the peeps data from the database
+    #display the peeps int he view
+    @peeps = Peeps.all
+    erb :peeps
+  end
+
+  post '/add_peep' do
+    Peeps.add_peep(message: params['message'])
+    redirect '/peeps'
```
noravdh: I think there's an opportunity here to separate the redirect into it's own GET request :)

spec/features/chitter_feature_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    expect(page).to have_content "Test peep1"
+  end
+
+  scenario 'it can add a peep' do
+    visit ('/peeps')
+    fill_in :message, with: 'I have added a peep'
+    click_button 'submit'
+    expect(page).to have_content('I have added a peep')
+  end
+end
```
noravdh: really well-built tests here! It could be worth separating each feature into its own file to keep the file from becoming too complex as you add more features.

spec/units/chitter_unit_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+
+  it 'can add a peep' do
+    Peeps.add_peep(message: 'I have added a peep')
+
+    # Add the new peep data
+    expect(Peeps.all).to include 'I have added a peep'
+
+  end
```
noravdh: great tests, & perfect test coverage 🎉

app.rb
```diff
@@ -1,8 +1,21 @@
 require 'sinatra/base'
+require './lib/peeps'
 
 class Chitter < Sinatra::Base
-  get '/test' do
-    'Test page'
+  get '/' do
+    'Welcome to Chitter!'
+  end
```
noravdh: From a user journey perspective, could it be worth adding some functionality to the homepage? Just so people know to go to /peeps? Or maybe all the functionality of /peeps could sit on the homepage instead?

PR #11, opened 2020-03-17
Gary chitter

lindseydew:


lib/peeping.rb
```diff
@@ -0,0 +1,18 @@
...
+		@date = date
+	end
+	def self.all 
+		if (ENV["ENVIRONMENT"] == 'test')
+			connection = PG.connect(dbname: 'chitter')
+		else 
+			connection = PG.connect(dbname: 'chitter_test')
+		end
+		result = connection.exec("SELECT * FROM peeps;")
+		result.map { |tweet| Tweet.new(message: tweet['message']) }
```
lindseydew: Excellent!

lib/peeping.rb
```diff
@@ -0,0 +1,18 @@
...
+
+class Tweet
+	attr_reader :message, :date
+	def initialize(message, date = '')
+		@message = message
+		@date = date
+	end
+	def self.all 
+		if (ENV["ENVIRONMENT"] == 'test')
+			connection = PG.connect(dbname: 'chitter')
```
lindseydew: Nice one for checking the test and the real db here

spec/features/peeping_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+
+describe Tweet do 
+
+  it 'it views all peeps' do 
+		expect(Tweet.all.class). to eq Array
+  end
+  
+	it 'creates peeps' do 
+		tweet = Tweet.new('Im peeping', '14.11.1988')
+		expect(tweet.message).to eq 'Im peeping'
```
lindseydew: Excellent unit test here!

spec/features/peeping_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'peeping'
+
+describe Tweet do 
+
+  it 'it views all peeps' do 
+		expect(Tweet.all.class). to eq Array
```
lindseydew: This is great start. 
Perhaps one way of checking the `all` method could be to insert a row into the database at the beginning of this test and then check that the method returns an array of a Tweet with the same attributes as in the database?

spec/features/test_page_spec.rb
```diff
@@ -3,4 +3,8 @@
     visit('/test')
     expect(page).to have_content "Test page"
   end
+  scenario 'page says hello chitter' do
+    visit('/')
+    expect(page).to have_content "Chitter"
```
lindseydew: Similarly to testing the all method, this test could insert a row to the db at the beginning and then check the attributes of the tweet appear on the page

view/index.erb
```diff
@@ -0,0 +1,10 @@
+lets peep!
+<div>
+  <% if @tweets.length %>
+    <ul>
+      
+      <li><%= @tweet %></li>
```
lindseydew: Nice, totally on the right lines here. What do you see if you run your app manually? 

I think if you want to display each tweet, would need to iterate over the `@tweets` variable. Similar to this: 
https://github.com/makersacademy/mvc-example/blob/master/views/books.erb#L13

PR #17, opened 2020-05-22
first commit for pull request

george-bit:


george-bit:
Morning,

So I managed to implement the code following the first 2 user stories without too much issue. The part I got stuck on was the implementation of date time. My application code is functioning in the way I want it to currently but according to my failing test this returns '1993' instead of the date. On review, I realised that this is because the datatype on my database column is VARCHAR and it was taking the string of the date passed to it '2020-05-22' it was executing this as an equation!

With a little more work I think I could make this functional.

Thanks
George

PR #19, opened 2020-05-22
Anna's chitter

annamvoitika:


lib/peep.rb
```diff
@@ -0,0 +1,43 @@
...
+    Peep.new(message: result[0]['message'], created_at: result[0]['created_at'])
+  end
+
+  def self.find(keyword:)
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
+    result = connection.exec("SELECT message FROM peeps WHERE message LIKE '%#{keyword}%';")
```
JasmineRachel, line 38: What does the % do?

app.rb
```diff
@@ -1,8 +1,34 @@
 require 'sinatra/base'
+require './lib/peep.rb'
```
JasmineRachel, line 3: Good use of RESTful routes! :)

PR #21, opened 2020-05-23
Kabir's Pull Request

twizlar12:


views/peeps/index.erb
```diff
@@ -0,0 +1,9 @@
+<ul>
+  <% @peeps.each do |peep| %>
+    <li><%= peep %></li>
+  <% end %>
+</ul>
+
+<form method="get" action="/">
+    <button type="submit">Return to Homepage</button>
```
cass-93, line 8: Like the use of button to return to the home page rather than a link which is something I didn't use

db/migrations/02_add_date_to_peeps.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE peeps ADD COLUMN date DATE;
```
cass-93, line 1: Also the use of the DATE syntax to store the current date is something I didn't implement either and something I can take away from this, good job!

PR #22, opened 2020-05-23
First User Story completed, feature test, unit test both passing, pro…

JonF1:
…ject set up

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,8 +1,18 @@
 require 'sinatra/base'
+require './lib/message.rb'
 
 class Chitter < Sinatra::Base
+  get '/' do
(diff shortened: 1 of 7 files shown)


JonF1:
The first three user stories are included in the pull request. I'd like to improve my refactoring so if there is anything you can see for that, that would be great! Thanks!

Adewunmi12:
I really liked the way you laid out your codes. It was quite easy to read and it shows you understood the necessary steps you had to take in order to achieve the user stories. You should be really proud on how far you got with the challenge.

As you mentioned about the refactoring, not sure really what you can change to make more simplified as this is an area I am struggling with also.

PR #23, opened 2020-05-26
Glen Commit

GlenHani:
Glen Done

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,9 +1,27 @@
 require 'sinatra/base'
+require './lib/peeps'
+require './lib/db_connection'
 
 class Chitter < Sinatra::Base
   get '/test' do
     'Test page'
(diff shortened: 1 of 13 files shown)


george-bit:
Hi Glen,

Looking over your code I'm really impressed it looks dry and you're implemented OOP on your main Ruby code. In particular, the encapsulation of the Database connection is good practice and I will personally make note of this for my own project.

The only notes I have are minor. Your views could be a bit more user friendly but that's easy to remedy. You need to enable sessions on your app.rb file as discussed. Also, look into the datatpye of your SQL new column, to achieve the desired effect you'll need to use TIMESTAMP as opposed to CURRENT DATE.

Other than that really clean and well implemented code. Well done!

Many thanks
George

PR #24, opened 2020-05-26
got stuck at user story 3

JasmineRachel:


app.rb
```diff
@@ -1,9 +1,27 @@
...
   end
 
+  get '/messages' do
+    @messages = Messages.all
+    erb :'messages/index'
+  end
+
+  get '/messages/new' do
+    erb :'messages/new'
+  end
```
annamvoitika, line 18: I love that you used RESTful routes!

lib/messages.rb
```diff
@@ -0,0 +1,35 @@
...
+    if ENV['ENVIRONMENT'] == 'test'
+    connection = PG.connect(dbname: 'chitter_test')
+    else
+    connection = PG.connect(dbname: 'chitter')
+    end
+    result = connection.exec("INSERT INTO peeps (message, date_posted) VALUES('#{message}', '#{date_posted}') RETURNING message, date_posted")
+    # Messages.new(message: result['message'], date_posted: result['date_posted'])
+   end
+
+ end
```
annamvoitika, line 35: Your code is very easy to read and follow! The only improvement I could think of is removing the commented out code for the final version, but apart from that I think there is nothing else I could really suggest. Well structured, clear, indented and concise code! :)

PR #25, opened 2020-05-26
User story one and two done

Adewunmi12:
So I was only able to complete the first and second user stories. 

I found the challenge ok but it helped me understand what I need to work on.

### DIFF
--- a/lib/comments.rb
+++ b/lib/comments.rb
@@ -1,0 +1,22 @@
+require 'pg'
+
+class Comments
+
+  attr_reader :message
+
+  def initialize(message:)
+    @message = message
+  end
+
+  def self.all
+    result = DatabaseConnection.query("SELECT * FROM peeps;")
+    result.map do |peep|
+      Comments.new(message: peep['message'] )
+    end
+  end
+
+  def self.create(message:)
+    result = DatabaseConnection.query("INSERT INTO peeps (message) VALUES('#{message}') RETURNING message")
+    Comments.new( message: result[0]['message'] ) 
+  end
+end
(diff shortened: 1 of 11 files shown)


lib/database_connection.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  def self.connection
+    @connection
+  end
+
+  def self.query(sql)
+    @connection.exec(sql)
+  end
+end
```
JonF1, line 15: I thought this was great, having a separate class to do the database connection work! It makes the Comments class a lot neater. Great idea!

spec/features/adding_comment_spec.rb
```diff
@@ -0,0 +1,9 @@
+feature 'adding message' do
+  scenario 'add a comment' do
+    visit '/new'
+    fill_in('message', with: "adfa")
+    click_button('Submit')
+
+    expect(page).to have_content "adfa"
+  end 
+end
```
JonF1, line 9: You may be able to use the test database for this feature test here

lib/comments.rb
```diff
@@ -0,0 +1,22 @@
...
+    result.map do |peep|
+      Comments.new(message: peep['message'] )
+    end
+  end
+
+  def self.create(message:)
+    result = DatabaseConnection.query("INSERT INTO peeps (message) VALUES('#{message}') RETURNING message")
+    Comments.new( message: result[0]['message'] ) 
+  end
+end
```
JonF1, line 22: Really neat model here!

JonF1:
I thought this was great and I particularly liked the seperation of concerns in your model classes, that's definitely something that I should have done.

The only thing I can pick up on is whether you're tests are always accessing the test database when they are being run. In terms of the third user story you could have altered the table to take in a date input. 

But apart from that, really easy to read and neat code, really well done!

PR #28, opened 2020-11-06
Jack Morris Chitter Challenge

jckmrrs:


spec/features/search_spec.rb
```diff
@@ -0,0 +1,18 @@
+require_relative 'web_helpers'
```
swizelc, line 1: nice use of web helpers :)

PR #40, opened 2020-11-09
My code

christian-naykene:


lib/messages.rb
```diff
@@ -0,0 +1,31 @@
...
+class Message
+
+    attr_reader :chit
+
+    def initialize(chit:)
+        @chit = chit
+    end
+
+    def self.list
+        if ENV['ENVIRONMENT'] == 'test'
```
FabihaHassan123, line 12: As setting up the connection is repeated through methods you can create a private method or a class which might be better.

views/view_messages.erb
```diff
@@ -0,0 +1 @@
+<%= @messages %>
```
FabihaHassan123, line 1: over here you can just call message.chit in an each loop to view just the chits.

FabihaHassan123 (commented):
Your code was very clean and organised which made it easy to read and follow. Just left a couple comments to help you keep your code DRY.

PR #45, opened 2021-02-19
chitter challenge user stories 1 and 2

girl-interrupted:


app.rb
```diff
@@ -1,9 +1,24 @@
 require 'sinatra/base'
+require './lib/peep'
 
 class Chitter < Sinatra::Base
   get '/test' do
     'Test page'
   end
 
+  get '/' do
```
EdwardAndress, line 9: Nice, RESTful routes.

lib/peep.rb
```diff
@@ -0,0 +1,25 @@
+require 'pg'
+
```
EdwardAndress, line 2: watch out for the indentation! Rubocop would not like this 🤖

EdwardAndress:
This is a nice solution.  How would you add another test to ensure that the peeps are displayed in the correct order?

PR #46, opened 2021-02-21
Chitter Challenge Complete

yy140:


app.rb
```diff
@@ -1,9 +1,33 @@
...
+  get '/peeps/new' do
+    erb :'peeps/new'
+  end
+
+  post '/peeps' do
+    Peeps.create(message: params[:message])
+    redirect '/peeps'
+  end
+  
+  get '/peeps/search' do
```
EdwardAndress, line 27: can you think of a way to combine `get /peeps/search` and 'get /peeps'?

lib/peeps.rb
```diff
@@ -0,0 +1,48 @@
...
+  attr_reader :id, :date, :message
+
+  def initialize(id:, date:, message:)
+    @id = id
+    @date = date
+    @message = message
+  end
+
+  def self.all
+
```
EdwardAndress, line 14: This is good, in that it works, but I'd probably move the logic that determines which db is used into another class.

lib/peeps.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def self.filter(keyword)
+
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
+    result = connection.exec("SELECT * FROM peeps WHERE message LIKE ('%#{keyword}%')")
```
EdwardAndress, line 45: nice SQL!

EdwardAndress:
Nice work, Anna!  How could you add another test to ensure that the peeps are displayed in the correct order?

PR #47, opened 2021-02-21
chitter challenge

JFindley95:
makers end of module challenge for database week.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,8 +1,20 @@
 require 'sinatra/base'
+require './lib/peep'
 
 class Chitter < Sinatra::Base
-  get '/test' do
-    'Test page'
+
+  get '/' do
+    erb :homepage
+  end
+
+  get '/peeps' do
+    @peeps = Peep.all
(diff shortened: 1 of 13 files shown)


app.rb
```diff
@@ -1,8 +1,20 @@
...
+  get '/' do
+    erb :homepage
+  end
+
+  get '/peeps' do
+    @peeps = Peep.all
+    erb :view
+  end
+
+  post '/create_peeps' do
```
EdwardAndress, line 15: to be RESTful, this should be `post "/peeps"`

lib/peep.rb
```diff
@@ -0,0 +1,22 @@
+require 'pg'
+
```
EdwardAndress, line 2: watch out for your indentation here!

EdwardAndress:
It's a good attempt! Well done :) I particularly like the styling 😎 

How could you write a test to ensure that the peeps are displayed in the correct order?

PR #48, opened 2021-02-21
Completed Chitter Challenge

vinc3108:
I need to work on my tests for the last 2 user stories and work on the design for the views.


app.rb
```diff
@@ -1,9 +1,29 @@
 require 'sinatra/base'
+require './lib/chitter'
 
 class Chitter < Sinatra::Base
   get '/test' do
     'Test page'
   end
```
EdwardAndress, line 8: nice RESTful routes!  Did you consider combining `get /peeps` and `get /peeps/matches`?

EdwardAndress: how could you do that?

EdwardAndress:
Nice work, Vince!  You've demonstrated the important things from this week 👍 

If you were to add another test that ensures that the peeps are displayed in the correct order, what would that look like?

PR #49, opened 2021-02-22
Chitter challenge

stacy1999:


spec/features/viewing_peeps_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'pg'
+
+feature 'Viewing peeps' do
+  scenario 'visiting the index page' do
+    
+    Peep.create(message: "None of your emails are finding me well" )
```
EdwardAndress, line 6: 😆

EdwardAndress:
Nice work!  I think you've demonstrated the important things from this week.  If you were to add another test to ensure that the peeps are shown in the correct order, how would you do that?

PR #50, opened 2021-02-22
Main

HarryCarton:


app.rb
```diff
@@ -1,8 +1,52 @@
...
+    @peeps = PeepsManager.all
+    erb(:all_peeps)
+  end
+
+  get '/create_peep' do
+    erb(:create_peep)
+  end
+
+  post '/peeps' do
+    #peep = Peep.new(params[:message], session[:user])
```
EdwardAndress, line 37: Looks like you were trying to associate a peep with a user here.  How did it go?  DM me if you want some tips!

app.rb
```diff
@@ -1,8 +1,52 @@
...
-  get '/test' do
-    'Test page'
+
+  enable :sessions
+
+  get '/' do
+    erb(:sign_in)
+  end
+
+  post '/sign_in_conf' do
```
EdwardAndress, line 14: nice work with sign in!

If you want to check that a user is signed in before all controller actions, look up 'sinatra controller hooks' and you'll find a way of executing some code before any controller action (and you can check for a user in the session there)

EdwardAndress:
Really nice work, Harry!  The next step is to add some hooks to the controller so that anyone who is not signed in is directed to the sign in page.

EdwardAndress:
Oh and, you should look into password encryption so that your user's passwords are safe! :)

PR #51, opened 2021-02-22
Week 6: Databases - Chitter Project

RowHunt:


app.rb
```diff
@@ -1,9 +1,15 @@
...
 class Chitter < Sinatra::Base
   get '/test' do
     'Test page'
   end
 
+  get '/show' do
+    @peeps = Peep.all
+    erb :index
+  end
+
```
EdwardAndress, line 13: I see you implemented the `Peep.create` method.  The controller action for that would look something like this
```
post '/peeps' do
  Peep.create(params[:peep])
  redirect '/peeps'
end
```

depending on what the form for creating peeps looks like (and how the params are structured).

EdwardAndress:
You've made a good start here and demonstrated a lot of the important stuff from this week.  The only thing you're missing is being able to create peeps by using a form.  If you like we can look at that together sometime – DM me if you that would be helpful!

PR #52, opened 2021-02-23
Chitter Challenge

harrythomson1:
I have mixed feelings about this one.

Really happy that I was able to keep everything on the same page within the browser.

Happy I was able to implement all the user stories.

Happy I was able to understand how a database worked and how to access data.

I'm not happy with it though as I feel like I have heavily borrowed a lot of the code from the resources that we had been using throughout the week.

Also didn't style it but this was more of a time thing as I was focused on completing the technical parts of the challenge before attaching a style sheet.


app.rb
```diff
@@ -1,8 +1,30 @@
...
+require './lib/peep'
+require './database_connection_setup'
 
 class Chitter < Sinatra::Base
-  get '/test' do
-    'Test page'
+
+  enable :sessions
+
+  get '/peeps' do
```
EdwardAndress, line 9: Nice work integrating the search functionality into this route!

Looks a bit like you're going to filter the peeps in the view.  You could also do it here and then pass @peeps, filtered or not, into the view.

EdwardAndress: that would mean you only need one view template

app.rb
```diff
@@ -1,8 +1,30 @@
...
+    @peeps = Peep.all
+    @filter = session[:filter]
+    if @filter.nil?
+      erb :peeps
+    else
+      erb :filtered_peeps
+    end
+  end
+
```
EdwardAndress, line 18: to be RESTful, this route should be `post /peeps`

app.rb
EdwardAndress: you could also create a new method on the peep class to filter the results at the DB level

EdwardAndress:
This is a really nice attempt! I particularly like the way you're handling 'search' inside `get /peeps`.  You can go a bit further though – see comments in line against the code.

PR #56, opened 2021-04-06
Chitter Challenge

al01010:
Can see peeps in reverse order, based on date - but not on time.


views/peeps/index.erb
```diff
@@ -0,0 +1,7 @@
+<ul>
+  <% @peeps.each do |peep| %>
+    <li><%= peep.message %> <%= peep.date%> </li> 
+  <% end %>
+</ul>
+
+<a href="/peeps/new">Add Peep</a>
```
Maelvis, line 7: I like that there is a link to the 'new' page on the index page! Makes it very accessible to the user.

app.rb
```diff
@@ -1,9 +1,32 @@
...
+    @peeps = Peep.all
+    erb :'peeps/index'
+  end
+
+  get '/peeps/new' do
+    erb :'peeps/new'
+  end
+
+  post '/peeps' do
+    @peep_date = Time.new.strftime("%Y-%d-%m")
```
Maelvis, line 25: I like this way of entering the time to the DB. You could potentially also make the DB enter the time/date automatically when you create a new row (peep)

lib/peep.rb
```diff
@@ -0,0 +1,37 @@
...
+
+class Peep
+  attr_reader :message, :date
+
+  def initialize(message:, date:)
+    @message = message
+    @date = date
+  end
+
+  def self.all
```
Maelvis, line 11: You could perhaps extract the code that checks if you are in a test environment to another file in a function and just reference it here, to avoid repeating.

spec/features/viewing_peeps_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    visit('/peeps')
+
+    expect(page.find('li:nth-child(1)')).to have_content 'I like cheese!'
+    expect(page.find('li:nth-child(2)')).to have_content 'This is the third peep!'
+    expect(page.find('li:nth-child(3)')).to have_content 'This is another peep!'
+    expect(page.find('li:nth-child(4)')).to have_content 'This is a peep!'
+    
+  end
+
+end
```
Maelvis, line 39: Very clear tests and great test coverage :)

PR #59, opened 2021-04-06
PH Chitter Challenge

Phali18:


Dominic-Taylor-Dev:
Really great effort. I don't have anything to add to it as it's quite different to where I got up to, and I don't understand enough to comment. There are a few things that wouldn't have occurred to me (I didn't finish last week's pair programming task so they may have been the 'proper' way to do things). These include:

1) Breaking out the database connection into its own class (and I didn't actually understand  what you were doing with this: class << self; attr_reader :connection; end
2) Using Bootstrap to make it all visually nicer (although I think that's a Twitter thing originally, so I guess it makes sense...!)
3) Using CSS similarly - although I will almost certainly try to copy you a bit here :)
4) All of the stuff about authentication and passwords (nice!)
5) The RESTful routes (I did actually know about this but just didn't really implement it myself in the way that you did - you put me to shame here...!)

The one problem I did have was trying to make it run locally. bundle install worked fine, but then actually trying to run app.rb didn't work for me because of issues in the 'database_connection_set_up' file and dependencies. May just be a minor teething issue (or something to spell out more in the README) but meant I couldn't check out the working app.

PR #65, opened 2021-05-21
chitter-challenge-liam

lkilk:
4 user stories completed


database_connection_setup.rb
```diff
@@ -0,0 +1,7 @@
+require './lib/database_connection'
+
+if ENV['ENVIRONMENT'] == 'test'
+  DatabaseConnection.setup('chitter_test')
+else
+  DatabaseConnection.setup('chitter')
+end
```
adamroy12, line 7: This is really cool. Nicely refactored.

spec/database_helpers.rb
```diff
@@ -0,0 +1,7 @@
+require 'pg'
+
+def persisted_data(id:)
+  connection = PG.connect(dbname: 'chitter_test')
+  result = connection.query("SELECT * FROM peeps WHERE id = #{id};")
+  result.first
+end
```
adamroy12, line 7: I didn't know if I needed the persisted data/id stuff. Since we don't have to add a feature for updating peeps, the only thing we'd use it for is testing that the peeps we've added to the database have the same id as the ones we return from the database. And we can test which peeps get returned against our expectations just by comparing the string. It doesn't hurt, but I don't know if it's a useful test in this situation.

spec/peep_spec.rb
```diff
@@ -1,22 +1,34 @@
...
+      Peep.create(message: 'this is my first peep!')
 
       peeps = Peep.all
-      expect(peeps).to include('This is a peep!')
+
+      expect(peeps.length).to eq 2
+      expect(peeps.first).to be_a Peep
+      expect(peeps.first.id).to eq peep.id
+      expect(peeps.first.message).to eq 'This is a peep!'
+      expect(peeps.first.posted_on).to eq peep.posted_on
```
adamroy12, line 19: So you'd know if the first peep returned was the same one you added first because it'll be the only one with the string message 'This is a peep!'. So, I don't know if the id test is useful, it seems like we're testing the same thing twice. It is possible it could be useful though, it just feels superfluous.

spec/features/view_peeps_spec.rb
```diff
@@ -10,10 +11,21 @@
...
+  # So that I can easily see the latest peeps
+  # I want to see a list of peeps in reverse chronological order
+
+  feature 'lists latest peeps first' do
+    scenario 'list of peeps in reverse chronological order' do
+      add_peep('This is my first peep')
+      add_peep('This is my latest peep')
+
+      expect(page.body.index('This is my latest peep')).to be < page.body.index('This is my first peep')
+    end
```
adamroy12: I couldn't work out how exactly to test if something comes before or after something else. This is really succinct way of doing it. Nice.

spec/database_helpers.rb
```diff
@@ -5,3 +5,9 @@ def persisted_data(id:)
   result = connection.query("SELECT * FROM peeps WHERE id = #{id};")
   result.first
 end
+
+def add_peep(message)
+  visit('/peeps/new')
+  fill_in('message', with: message)
+  click_button('Post')
+end
```
adamroy12, line 13: very nice. like this a lot

adamroy12 (commented):
Overall, really nice code. I especially liked the way you refactored a lot of the repetitive code, like connecting to databases, and adding peeps, into their own succinct methods you could just call whenever you needed them. Really cool, and I'll definitely be doing more of that myself. 
Again, I could be wrong about needing to test the id of elements of the databases in this instance. It did look like by testing them, some tests may be testing essentially the same things for the purposes of what we were doing. Like, whether a particular datum is returned from the database, which we could also know from testing which message string was returned (so long as there were no duplicates), which was something we were already testing. 
Not entirely sure, but I suppose it could help in case you're writing code for tests you don't need to be doing. Definitely could cut some unnecessary work out for you, and allow you to spend time doing other stuff.

PR #73, opened 2021-06-25
User Stories

Ryan734:
User Stories Review

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -79,6 +79,7 @@ GEM
       nokogiri (~> 1.8)
 
 PLATFORMS
+  x86_64-darwin-19
   x86_64-darwin-20
 
 DEPENDENCIES
--- a/README_USER_STORIES.md
+++ b/README_USER_STORIES.md
@@ -1,0 +1,25 @@
+As a Maker
+I want to see all the messages (peeps)
+in a browser
+So that I can see what people are doing
+
+
+As a Maker
+I want to post a message (peep) to chitter
+So that I can let people know what I am doing  
+
+
+As a Maker
+I want to see the date the message was posted
+So that I can see when people are doing things
+
+(Hint the database table will need to change to store the date too)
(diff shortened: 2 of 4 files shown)


Ryan734:
Second stage update.

app.rb
```diff
@@ -1,9 +1,31 @@
...
   end
 
+  get '/' do
+    @peeps = Peep.all
+    erb :peeps
+  end
+
+  post '/' do
+      message = params['peep']
+      connection = PG.connect(dbname: 'chitter_test')
```
Aroax, line 21: is it worth perhaps moving this DB connection to it's own object (class)? Rather than having it publicly exposed in the controller?

lib/peep.rb
```diff
@@ -0,0 +1,14 @@
+require 'pg'
+
+class Peep
+  def self.all
+    if ENV['ENVIRONMENT'] = 'test'
```
Aroax, line 5: could reference that external database object again here, would help DRY code up a bit as well as encapsulate

spec/features/viewing_peeps_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+    connection.exec("INSERT INTO peeps VALUES(1, 'hi there Chitter friends');")
+    connection.exec("INSERT INTO peeps VALUES(2, 'Chits for all');")
+    connection.exec("INSERT INTO peeps VALUES(3, 'Come and join the party');")
+
+    visit('/')
+
+    expect(page).to have_content "Welcome to the Chitter site"
+    expect(page).to have_content "hi there Chitter friends"
+    expect(page).to have_content "Chits for all"
+    expect(page).to have_content "Come and join the party"
```
Aroax, line 14: really clean and well laid out test, very easy to understand, love this approach!

Aroax (commented):
Great work so far!

PR #74, opened 2021-06-25
Emily Fitzs Chitter Challenge

EmilyFitzs:


lib/tweets.rb
```diff
@@ -0,0 +1,67 @@
...
+  attr_reader :id, :message, :user_id
+
+  def initialize(id:, message:, user_id:)
+    @id = id
+    @message = message
+    @user_id = user_id 
+  end
+
+def self.all
+  if ENV['ENVIRONMENT'] == 'test'
```
Jestfer, line 14: It's worth considering that this if/else statement is repeated below as well, to follow the DRY principle, you could, for example, create a function that takes the `environment` information and then reuse it to establish the connection with the test or the main database.

lib/tweets.rb
```diff
@@ -0,0 +1,67 @@
...
+  else
+    connection = PG.connect(dbname: 'chitter')
+  end
+
+  result = connection.exec("INSERT INTO peeps(message, user_id) VALUES('#{message}', #{user_id}) RETURNING id, message, user_id")
+  Tweets.new(id: result[0]['id'], message: result[0]['message'], user_id: result[0]['user_id'])
+end
+end
+
+class User
```
Jestfer, line 37: I would move this `User` class to its own file, i.e.`user.rb`

spec/web_helper.rb
```diff
@@ -0,0 +1,7 @@
+def sign_up(email: 'test@test.com', password: 'password', password_confirmation: 'password' )
```
Jestfer, line 1: Nice use of helpers!

lib/tweets.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+  result = connection.exec("SELECT * FROM peeps")
+  result.map do |user|
+    User.new(password: user['password'], email: user['email'])
+  end
+end
+
+  def password=(password)
+    @password = password
+    self.password_digest = BCrypt::Password.create(password)
```
Jestfer, line 56: Really great work on encrypting the passwords before being stored onto the database!

Jestfer:
Nice work on this! You have partially demonstrated the learning objectives for this unit, by completing the first two user stories. Well done on your submission, there's a good demonstration of defining both GET and POST routes, passing information from View -> Controller -> Model and back again.

If you decided to come back to this, a good challenge to try out would be to refactor the `Tweets` class file and extract the `User` class to its own file. Then, I'd work on the third user story to add the date to the peeps, and finally the 2 latest user stories, which are more meant as a bonus.

Finally, really nice work on the User authentication overall! Would you say part of your objectives for this challenge was to consolidate that section perhaps from MakersBnB two weeks ago? If so, that's great, but it would probably help you in any future tasks to follow the requirements specified in the user stories and add some nice extras (like user authentication in this case, which was not mandatory) if you have the time. Probably the fact that you decided to work on user authentication took a good amount of your Friday afternoon and prevented you from working on other user stories.

Well done! Let me know if you have any questions I can help with.

PR #75, opened 2021-06-25
Chitter challenge

ciombrik:


lib/chitter.rb
```diff
@@ -0,0 +1,20 @@
+require "pg"
+
+class Chitter
+  def self.all
+    if ENV["ENVIRONMENT"] == "test"
```
Jestfer, line 5: It's worth considering that this if/else statement is repeated below as well, to follow the DRY principle, you could, for example, create a function that takes the `environment` information and then reuse it to establish the connection with the test or the main database.

lib/user.rb
```diff
@@ -0,0 +1,34 @@
...
+class User
+  attr_reader :id, :email
+  def initialize(id:, email:)
+    @id = id
+    @email = email
+  end
+
+  def self.create(email:, password:)
+    encrypted_password = BCrypt::Password.create(password)
+    if ENV["ENVIRONMENT"] == "test"
```
Jestfer, line 12: Same as above! And probably now the best idea is to extract that function to a helper that you can use in both classes you decided to implement!

lib/user.rb
```diff
@@ -0,0 +1,34 @@
...
+
+class User
+  attr_reader :id, :email
+  def initialize(id:, email:)
+    @id = id
+    @email = email
+  end
+
+  def self.create(email:, password:)
+    encrypted_password = BCrypt::Password.create(password)
```
Jestfer, line 11: Really great work on encrypting the passwords before being stored onto the database!

Jestfer:
Nice work on this! You have partially demonstrated the learning objectives for this unit, by completing the first two user stories. Well done on your submission, there's a good demonstration of defining both GET and POST routes, passing information from View -> Controller -> Model and back again.

If you decided to come back to this, a good challenge to try out would be to add more tests to the existing application, as it seems some key tests are missing (e.g. `Chitter` class - create peep functionality, related with the 2nd functionality asked for in the user stories). Also, adding some feature tests with the main functionality that is asked for as part of the challenge would be great. Then, I'd work on the third user story to add the date to the peeps, and finally the 2 latest user stories, which are more meant as a bonus.

Finally, really nice work on the User authentication overall! Would you say part of your objectives for this challenge was to consolidate that section perhaps from MakersBnB two weeks ago? If so, that's great, but it would probably help you in any future tasks to follow the requirements specified in the user stories and add some nice extras (like user authentication in this case, which was not mandatory) if you have the time. Probably the fact that you decided to work on user authentication took a good amount of your Friday afternoon and prevented you from working on other user stories.

Well done! Let me know if you have any questions I can help with.

PR #76, opened 2021-06-28
Chitter Challenge

mjconn:


app.rb
```diff
@@ -1,9 +1,41 @@
...
+  get '/' do
+    @peeps = Peep.all
+    erb(:index)
+  end
+
+  get '/new_peep' do
+    erb(:new_peep)
+  end
+
+  post '/post_peep' do
```
emibrand, line 27: Nice RESTful routes!

spec/units/peep_spec.rb
```diff
@@ -0,0 +1,18 @@
+require_relative '../setup_test_database.rb'
+
+describe Peep do
+
+  it 'shows all peeps' do
+    add_row_to_test_database
```
emibrand, line 6: Nice use of a test helper

emibrand:
Really great work Mike! How could you add a sign in and out feature?

README.md
```diff
@@ -1,66 +1,11 @@
...
 
-* Challenge time: until the end of the day
-* Feel free to use google, your notes, books etc but please work on your own
-* Please raise a pull request when you start this challenge, and keep pushing updates as and when you make commits throughout the day
-* There is _no expectation_ to finish all or any of the user stories, please use this time to reflect on where you feel you are with the skill and what may support your learning.
-* If you get blocked, please reflect on what blocked you and any strategies you adopted that helped you make progress.
+*
 
-We are going to write a small Twitter clone that will allow the users to post messages to a public stream.
+## Database Setup
```
Jestfer, line 5: Great work on adding instructions for your challenge on the README, keep it up!

app.rb
```diff
@@ -1,9 +1,41 @@
...
+  require 'sinatra/reloader' if development?
+
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  enable :sessions
+  attr_reader :peep_date
+
   get '/test' do
```
Jestfer, line 14: It's good to remove any extraneous code from your implementation before making a Pull Request - there'll be lots of testing and exploration as you work on something, but it tidies it up to remove anything that is no longer needed for the program to meet its specification.

lib/peep.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def initialize(id, message, peep_date)
+    @id = id
+    @message = message
+    @peep_date = peep_date
+  end
+
+  def self.all
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
```
Jestfer, line 15: It's worth considering that this if/else statement is repeated below as well, to follow the DRY principle, you could, for example, create a function that takes the `environment` information and then reuse it to establish the connection with the test or the main database.

app.rb
```diff
@@ -1,9 +1,41 @@
...
+    @peeps = Peep.all
+    erb(:index)
+  end
+
+  get '/new_peep' do
+    erb(:new_peep)
+  end
+
+  post '/post_peep' do
+    time = Time.new
```
Jestfer, line 28: Normally, it's considered best practices to keep our controllers as short and concise as possible (i.e. `skinny controllers`). In this case, I would refactor this route and move the logic related with the `time` to the Model instead. How does this sound?

spec/features/view_peeps_spec.rb
```diff
@@ -0,0 +1,56 @@
+# As a Maker
```
Jestfer, line 1: It's great to have this sort of comments to guide your process. However, I'd remove these comments now as part of refactoring and cleaning up application before submitting.

spec/units/peep_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+    peeps = Peep.all
+    expect(peeps.first['message']).to eq 'This is a peep!'
+  end
+
+  it 'adds peeps' do
```
Jestfer, line 11: This test does not seem to test that your application can add a peep to the database, but it seems you're rather testing that you can establish a connection manually with the db via PG and, after using the helper to add a row to the test db, retrieving all of the results. 

But this is something that happens in the previous test where you test the `.all` function from the `Peep` class if I'm not mistaken.

Jestfer:
Nice work on this! You have demonstrated the learning objectives for this unit, by completing all of the user stories! Well done on your submission, there's a good demonstration of defining both GET and POST routes, passing information from View -> Controller -> Model and back again, as well as some good tests (this could be improved too! And I've added some more info where relevant).

If you decided to come back to this, a good challenge to try out would be to revisit your tests that interact with the database, especially for the POST. And if you want to keep working on Chitter, I'd suggest taking a look at your controller and Peep model again and trying to diagram what should belong in each of these. It's a nice rule of thumb to think about the controller "only" caring about the message it receives, passing it down to the model, and then sending it back after the model's done its work! Let me know if this helps! Hope it does.

Well done!

PR #77, opened 2021-06-28
Usna's Chitter challenge (incomplete)

usnahakimi:
Finished the second user story.


app.rb
```diff
@@ -1,9 +1,24 @@
...
+    erb :new
+  end
+
+  post '/messages' do
+    Message.create(message: params[:message])
+    redirect '/messages'
+  end
+
   run! if app_file == $0
 end
```
anitaRead, line 24: Great, well structured and easy to read code!

views/index.erb
```diff
@@ -0,0 +1,5 @@
+<ul>
+  <% @messages.each do |messages| %>
```
anitaRead, line 2: Great layout, you may want to name the iterator to the singular of messages.
E.g. 
@messages.each do |message|
message
end

lib/message.rb
```diff
@@ -0,0 +1,24 @@
...
+            connection = PG.connect(dbname: 'chitter_test')
+        else
+            connection = PG.connect(dbname: 'chitter')
+        end
+       
+        result = connection.exec("SELECT * FROM peeps;")
+        result.map { |message| message['message'] }
+    end
+
+    def self.create(message:)
```
anitaRead, line 15: Very well named class methods!

app.rb
```diff
@@ -1,9 +1,24 @@
 require 'sinatra/base'
+require './lib/message.rb'
 
 class Chitter < Sinatra::Base
   get '/test' do
```
Jestfer, line 5: It's good to remove any extraneous code from your implementation before making a Pull Request - there'll be lots of testing and exploration as you work on something, but it tidies it up to remove anything that is no longer needed for the program to meet its specification.

spec/messages_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+        expect(messages).to include("Hello, first message!")
+        expect(messages).to include("It is raining!")
+        expect(messages).to include("Today is Friday!")
+    end
+  end
+
+  describe '.create' do
+    it 'creates a new bookmark' do
+      Message.create(message: 'It is the weekend!')
```
Jestfer, line 22: How do we know in this test that the environment is `'test'` so that the right database is populated when we create a message? And think about the same when you retrieve the messages below as well.

lib/message.rb
```diff
@@ -0,0 +1,24 @@
+require 'pg'
+
+class Message
+    def self.all
+        if ENV['ENVIRONMENT'] == 'test'
```
Jestfer, line 5: This if/else statement is repeated below in the `.create` method as well. It's worth thinking about extracting this into its own method / helper.

Jestfer:
Nice work on this! You have partially demonstrated the learning objectives for this unit, by completing the first two user stories. Well done on your submission, there's a good demonstration of defining both GET and POST routes, passing information from View -> Controller -> Model and back again, as well as some good tests.

If you decided to come back to this, a good challenge to try out would be tackling the next user story. It would require some small changes to add the Time to your Model and to display it, but it makes for an interesting challenge when it comes to testing.

Well done!

PR #78, opened 2021-06-28
Chitter Challenge 

emibrand:


spec/database_helpers.rb
```diff
@@ -0,0 +1,7 @@
+require 'pg'
+
+def persisted_data(id:)
```
mjconn, line 3: Good use of this persisted data method

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
```
mjconn, line 2: Great test coverage

lib/peep.rb
```diff
@@ -0,0 +1,36 @@
...
+  attr_reader :id, :message, :date
+
+  def initialize(id:, message:, date:)
+    @id = id
+    @message = message
+    @date = date
+  end
+
+  def self.all
+    if ENV['ENVIRONMENT'] == 'test'
```
mjconn, line 14: Could put this if/else statement into a method of its own as used multiple times

mjconn (commented):
Looks great - clean code and good test coverage!

app.rb
```diff
@@ -1,9 +1,29 @@
 require 'sinatra/base'
+require_relative './lib/peep'
 
 class Chitter < Sinatra::Base
+  
   get '/test' do
     'Test page'
```
Jestfer, line 7: It's good to remove any extraneous code from your implementation before making a Pull Request - there'll be lots of testing and exploration as you work on something, but it tidies it up to remove anything that is no longer needed for the program to meet its specification.

lib/peep.rb
```diff
@@ -0,0 +1,24 @@
...
+      connection = PG.connect(dbname: 'chitter_test')
+    else 
+      connection = PG.connect(dbname: 'chitter')
+    end
+
+    result = connection.exec("SELECT * FROM peeps;")
+    result.map { |peep| peep['message'] }
+  end
+
+  def self.create(message:)
```
Jestfer: Well done! It's good Ruby practice to name a method like this '.create'

spec/units/peep_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'peep'
+require 'database_helpers'
+
+describe Peep do
+  describe '.all' do
+    it 'returns all peeps' do
+      connection = PG.connect(dbname: 'chitter_test')
+
+      peep = Peep.create(message: 'Peep 1', date: '2021-06-25')
```
Jestfer, line 9: I like the way that you have stored the first test case in a variable to check the assertions later on in a more convenient way.

spec/units/peep_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      Peep.create(message: 'Peep 2', date: '2021-06-25')
+      Peep.create(message: 'Peep 3', date: '2021-06-25')
+      
+      peeps = Peep.all
+
+      expect(peeps.length).to eq 3
+      expect(peeps.first).to be_a Peep
+      expect(peeps.first.id).to eq peep.id
+      expect(peeps.first.message).to eq 'Peep 1'
```
Jestfer, line 18: Related to my previous comment, I'd use the same approach here for the message (as you've done for the id): `...to eq peep.message`

And the same with the date as well. This is because you decided to store the first test case in a variable and this is what would make more sense as per your approach.

spec/units/peep_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  describe '.create' do
+    it 'creates a new peep' do
+      peep = Peep.create(message: 'Example Peep', date: '2021-06-25')
+      persisted_data = persisted_data(id: peep.id)
+
+
+      expect(peep).to be_a Peep
+      expect(peep.id).to eq persisted_data['id']
+      expect(peep.message).to eq 'Example Peep'
```
Jestfer, line 32: Same as discussed above.

spec/features/posting_a_peep_spec.rb
```diff
@@ -0,0 +1,10 @@
+feature 'Posting a peep' do
+  scenario 'A user can post a peep to Chitter' do
+    visit('/peep')
+    fill_in('message', with: 'Example Peep')
+    click_button('Submit')
+
+    expect(page).to have_content 'Example Peep'
+    expect(page).to have_content '2021-06-25'
```
Jestfer, line 8: What happens if I run this test today?

Jestfer: It seems the test is brittle because it's linked to a specific date (as per when the test was written). To avoid that, a suggestion would be to create a new time object and then format it to get the date, for example:

```ruby
time_today = Time.now
date_today = time_today.strftime("%Y-%m-%d")
```

Then you would expect the page to include the date as per today in this feature test.

Jestfer:
Nice work on this! You have demonstrated the learning objectives for this unit, by completing the first three user stories. Well done on your submission, there's a good demonstration of defining both GET and POST routes, passing information from View -> Controller -> Model and back again, as well as some good tests (this could be improved too! And I've added some more info where relevant).

If you decided to come back to this, a good challenge to try out would be to revisit your tests, especially when it comes to time. And if you want to keep working on Chitter, tackling the bonus User Stories after that would be a great idea!

Also, I'm curious to know why you named your third commit "Not working". What is it that is not working and that you would expect to work? It's a good practice to use descriptive commit messages to briefly explain what was part of that commit.

Well done!

PR #79, opened 2021-06-28
Irina's solo chitter

IrinaGC:


spec/peeps_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    it 'posts a new message(peep)' do
+      peep = Peeps.create(message: 'Hangover Sunday')
+      persisted_data = persisted_data(id: peep.id)
+
+      expect(peep).to be_a Peeps
+    
+      expect(peep.message).to eq 'Hangover Sunday'
+      expect(peep.date).to eq '2021-06-27'
+    end
+
```
Juel07, line 34: I like how you've segregated your tests for each method - makes it easier to read

lib/peeps.rb
```diff
@@ -0,0 +1,40 @@
...
+
+
+def self.create(message:)
+
+  if ENV['ENVIRONMENT'] == 'test'
+    con = PG.connect(dbname: 'chitter_test')
+  else
+    con = PG.connect(dbname: 'chitter')
+  end
+
```
Juel07, line 34: You could use `Time.new` which can give you the date & time - then that can be saved into a variable `time` which is then inserted into the DB along with the message.

spec/features/chitter_features_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    expect(page).to have_content "ToYoda is the coolest"
+  end
+
+  scenario 'to post a  new message (peep)' do
+    visit('/peeps/new')
+    fill_in('message', with: 'Hangover Sunday')
+    click_button('Peep')
+
+    expect(page).to have_content "Hangover Sunday"
+    expect(page).to have_content "2021-06-27"
```
penguat, line 24: Unfortunately it looks like this test no longer works, as we've passed the date it specified.

I'd suggest that for the future you pass the date to the database for these tests when you create the test entry - that way your tests would still be straightforward - and unlikely to break in the future.

While we're talking about the date field - you could help your fellow developers by including the ALTER TABLE statement you used to add the date column in the db/migrations folder. The db/migrations folder is a standard place for people to look for schema updates.

lib/peeps.rb
```diff
@@ -0,0 +1,40 @@
...
+  result = con.exec("INSERT INTO peeps (message) VALUES('#{message}') RETURNING id, message, posting_date;")
+  Peeps.new(id: result[0]['id'], message: result[0]['message'], date: result[0]['posting_date'])
+
+end
+
```
penguat, line 39: I really like this class - you've done all of this in a simple and elegant way, without unnecessary complexity

PR #80, opened 2021-06-28
Leslie's Chitter challenge

kabuk1:


lib/peep.rb
```diff
@@ -0,0 +1,24 @@
+require 'pg'
+require_relative 'database_connection'
+
```
goncalog, line 3: great way to setup the database connection!

lib/database_connection.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
+
+  def self.connection
+    @connection
+  end
+
+  def self.query(sql)
+    @connection.exec(sql)
+  end
+
```
goncalog, line 15: amazing work in extracting the DatabaseConnection and related methods from the Peep class - very good example of SRP principle

spec/database_connection_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'database_connection'
+
```
goncalog, line 2: this is minor but the require expressions from the spec files could be moved to the spec_helper file

spec/database_connection_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+describe DatabaseConnection do
+  describe '.setup' do
+    it 'sets up a connection to a database through PG' do
+      expect(PG).to receive(:connect).with(dbname: 'chitter_test')
+
```
goncalog, line 7: very elegant way to test the db connection

spec/peep_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  describe '.create' do
+    it 'creates a new peep' do
+      peep = Peep.create(message: 'I have arrived!')
+      persisted_data = persisted_data(id: peep.id)
+  
+      expect(peep).to be_a Peep
+      expect(peep.id).to eq persisted_data.first['id']
+      expect(peep.message).to eq 'I have arrived!'
+    end
+
```
goncalog, line 30: very thorough and well structured testing!

app.rb
```diff
@@ -1,9 +1,31 @@
...
+require 'pg'
+require './lib/peep'
+require './database_connection_setup'
 
 class Chitter < Sinatra::Base
+  configure :development do
+    register Sinatra::Reloader
+  end
+
   get '/test' do
```
eoinmakers, line 12: It's good to remove any extraneous or redundant code from your implementation, as a quick win for a cleanly written implementation.

eoinmakers, line 12: It's good to remove any extraneous or redundant code from your implementation, as a quick win for a cleanly written implementation.

spec/peep_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+      peep = Peep.create(message: 'This is a peep!')
+      Peep.create(message: 'Welcome to Chitter!')
+      Peep.create(message: 'My first peep!')
+
+      peeps = Peep.all
+
+      expect(peeps.length).to eq 3
+      expect(peeps.first).to be_a Peep
+      expect(peeps.first.id).to eq peep.id
+      expect(peeps.first.message).to eq 'This is a peep!'
```
eoinmakers, line 16: Really thorough unit testing of the Peep Model.

app.rb
```diff
@@ -1,9 +1,31 @@
...
     'Test page'
   end
 
+  get '/' do
+    erb :index
+  end
+
+  get '/peeps' do
```
eoinmakers, line 20: Really cleanly defined routes, with all of the business logic kept to the Peep model. Nice work on this.

spec/features/posting_peep_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+feature 'Posting a new peep' do
+  scenario 'User can post a peep to Chitter' do
+    visit('/')
+    fill_in('message', with: 'I have arrived!')
+    click_button('Submit')
+
+    expect(page).to have_content 'I have arrived!'
+  end
+
+  # scenario 'View the time message was posted' do
```
eoinmakers, line 12: As a first rendition of this test, you could define a timestamp with Time.new, create a new Peep, and then check that the Peep contains the expected information - timestamp.hour, .min, .day, etc. 

It's good practice to use stubs in place of information like Time (which is constantly moving), but it can be easier to implement it with real information first and then refactor as needed.

eoinmakers (commented):
Nice work! You have partially demonstrated the learning objectives for this unit of learning. Well done on achieving the first two user stories. You wrote them very cleanly, and there's a nice separation of concerns between Model, View and Controller in your implementation. 

I see you've made a start with the next user story, so if you decide to circle back to it, the unit testing is probably the most challenging part of implementing it. Do let me know if you have any questions about this review, my name is Eóin on Apprenticeships workspace!

Well done!

PR #81, opened 2021-06-28
Yusuf's attempt on Chitter

YAKhaliq:


app.rb
```diff
@@ -1,9 +1,30 @@
...
+require 'sinatra/reloader'
+require 'pg'
+require './lib/peep'
 
 class Chitter < Sinatra::Base
+  configure :development do
+    register Sinatra::Reloader
+  end
+
   get '/test' do
```
eoinmakers, line 11: It's good to remove any extraneous code from your implementation before making a Pull Request - there'll be lots of testing and exploration as you work on something, but it tidies it up to remove anything that is no longer needed for the program to meet its specification.

app.rb
```diff
@@ -1,9 +1,30 @@
...
     'Test page'
   end
 
+  get '/' do
```
eoinmakers, line 15: Instead of immediately redirecting to Peeps, you could try loading the peeps here instead. It will still meet the specification this way, as the Peeps will display on the homepage. This way, you would need one less route, which is slightly redundant in this instance.

lib/peep.rb
```diff
@@ -0,0 +1,31 @@
...
+      conn = PG.connect(dbname: 'chitter_test')
+    else
+      conn = PG.connect(dbname: 'chitter')
+    end 
+
+    result = conn.exec('SELECT * FROM peeps;')
+    result.map {|peep| Peep.new(peep['id'], peep['message'])}
+  end
+
+  def self.new(message)
```
eoinmakers, line 22: It's good Ruby practice to name a method like this '.create', rather than .new, or .add, for example.

spec/features/posting_peep_spec.rb
```diff
@@ -0,0 +1,7 @@
+feature 'Posting a peep' do
+  scenario 'User can post and view their peep' do
+    visit('/')
+    fill_in('peep', with: 'Hello World!')
+    expect(page).to have_content('Hello World!')
+  end
+end
```
eoinmakers, line 7: Another feature test you could potentially add here is testing that the homepage is viewable. This will test that the user can load the web app and they'll see everything they need in order to post their first Peep.

eoinmakers (commented):
Nice work on this! You have partially demonstrated the learning objectives for this unit, by completing the first two user stories. Well done on your submission, there's a good demonstration of defining both GET and POST routes, passing information from View -> Controller -> Model and back again, as well as some good tests. 

If you decided to come back to this, a good challenge to try out would be tackling the next user story. It would require some small changes to add the Time to your Model and to display it, but it makes for an interesting challenge when it comes to testing. 

Well done! Let me know if you have any questions about this, my name is Eóin on the Apprenticeships workspace.

PR #83, opened 2021-06-28
Hannah's chitter

hannah-vdw:
see readme for overview

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,8 +1,32 @@
 require 'sinatra/base'
+require './lib/peeps.rb'
 
 class Chitter < Sinatra::Base
+  enable :sessions
+
   get '/test' do
     'Test page'
+  end
+
+  get '/' do
+    @peeps = Peeps.all
+    erb :index
+  end
+
+  post '/addpeep' do
+    peep = params[:add_peep]
+    Peeps.add(peep)
+    redirect '/'
+  end
+
+  post '/search' do
+    session[:query] = params[:search]
+    redirect '/filter'
+  end
+
+  get '/filter' do
+    @peeps = Peeps.search(session[:query])
+    erb :filter
   end
 
   run! if app_file == $0
--- a/lib/peeps.rb
+++ b/lib/peeps.rb
@@ -1,0 +1,54 @@
+require 'pg'
+
+class Peeps
+
+  attr_reader :message, :date
+
+  def initialize(message:, date:)
+    @message = message
+    @date = date
+  end
+
+  def self.all
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
+
+    result = connection.exec('SELECT * FROM peeps ORDER BY id DESC;')
+    
+    result.map do |peep| 
+      Peeps.new(message: peep['message'], date: peep['date'])
+    end
+  end
+
+  def self.add(peep)
+    date = Time.new
+    store_date = "#{date.hour}:#{date.min}, #{date.day}.#{date.month}.#{date.year}"
+
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
+
+    result = connection.exec("INSERT INTO peeps (message, date) values ('#{peep}', '#{store_date}');")
+  end
+
+
+  def self.search(query)
+
+    if ENV['ENVIRONMENT'] == 'test'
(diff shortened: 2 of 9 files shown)


lib/peeps.rb
```diff
@@ -0,0 +1,54 @@
...
+    date = Time.new
+    store_date = "#{date.hour}:#{date.min}, #{date.day}.#{date.month}.#{date.year}"
+
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
+
+    result = connection.exec("INSERT INTO peeps (message, date) values ('#{peep}', '#{store_date}');")
```
EdwardAndress, line 36: Here, there are broadly two options in terms of how to achieve the search feature.

1. Retrieve all the peeps from the DB and then filter them in the controller or model, using Ruby.
2. Use SQL to only return the peeps of interest from the DB (which is the option you chose).

I'd say you made a good choice - nice work! Do you know why the one you chose is the better of these two options?

lib/peeps.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def self.search(query)
+
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
+
+    result = connection.exec("SELECT * FROM peeps WHERE message LIKE '%#{query}' OR message LIKE '%#{query}%' OR message LIKE'#{query}%';")
```
EdwardAndress, line 48: Here, there are broadly two options in terms of how to achieve the search feature.

1. Retrieve all the peeps from the DB and then filter them in the controller or model, using Ruby.
2. Use SQL to only return the peeps of interest from the DB (which is the option you chose).

I'd say you made a good choice - nice work! Do you know why the one you chose is the better of these two options?

spec/peeps_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'peeps'
+
+describe Peeps do
+
+  describe '.all' do
+    it 'returns a message from the db' do
+      connection = PG.connect(dbname: 'chitter_test')
+      connection.exec("INSERT INTO peeps (message) values ('This is a peep!');")
+      result = connection.exec('SELECT * FROM peeps;')
+      expect(result[0]['message']).to eq "This is a peep!"
```
EdwardAndress, line 10: You've implemented this feature nicely but I think the test could be improved a little because it would still pass even if your query only returned one, of the potentially many, peeps.

spec/peeps_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+      connection = PG.connect(dbname: 'chitter_test')
+      Peeps.add(new_peep)
+      result = connection.exec("SELECT * FROM peeps WHERE message = 'Hi everyone';")
+      expect(result[0]['message']).to eq "Hi everyone"
+    end
+  end
+
+  describe '.add' do
+    it 'adds a peep with a date' do
+      # need to stub in a date for the test
```
EdwardAndress, line 26: There's a gem called TimeCop that might help you here, though I'd have thought your test would be fine (most of the time) since the datestamp is accurate to minutes, rather than seconds.  I.e. You could probably do something like this...

```ruby
describe '.add' do  
    it 'adds a new peep to the db' do
      new_peep = "Hi everyone"
      datestamp = Time.new
      Peeps.add(new_peep)
      result = connection.exec("SELECT * FROM peeps WHERE message = 'Hi everyone';")
      expect(result[0]['date']).to eq "
          #{datestamp.hour}:#{datestamp.min}, #{datestamp.day}.#{datestamp.month}.#{date.year}
       "
    end
  end
```

Drop me a line on Slack if you want to chat about this!

app.rb
```diff
@@ -1,9 +1,33 @@
...
+require './lib/peeps.rb'
 
 class Chitter < Sinatra::Base
+  enable :sessions
+
   get '/test' do
     'Test page'
   end
 
+  get '/' do
```
EdwardAndress, line 11: You did a great job of implementing the stretch user stories – nice work!  If you had some time to refactor, I'd suggest rolling up your three controller actions (relating to generating lists of peeps) into one, here... something like this...

```ruby
get '/' do
  if params[:search]
    @peeps = Peeps.search(params[:search])
  else
    @peeps = Peep.all
  end
  erb :index
end
```

(You'd also have to refactor your search form)

EdwardAndress:
Nice work Hannah, you've clearly demonstrated the learning objectives for this week and done a nice job of implementing the hardest user story – filtering peeps.  I've left a few comments in line against your code.  Please drop me a line on Slack if you have any questions :)

PR #86, opened 2021-08-13
Lauren T-B

ltaylorbrown:


spec/messages_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+      setup_test_database
+      
+      Messages.create(message: 'Test message 1')
+      Messages.create(message: 'Test message 2')
+      Messages.create(message: 'Test message 3')
+
+      messages = Messages.all
+
+      expect(messages.length).to eq 3
+      expect(messages.first).to be_a Messages
```
ben-davis-devops, line 17: nice test to make sure that the database was responding as expected

views/new_message.erb
```diff
@@ -0,0 +1,4 @@
+<form action="/messages" method="post">
+  <input type="text" name="message" placeholder="New message" />
```
ben-davis-devops, line 2: Clear placeholder name so the user understands what should be entered.

lib/messages.rb
```diff
@@ -0,0 +1,37 @@
+require 'pg'
+
+class Messages
+attr_reader :id, :message, :date , :CURRENT_DATE
+
+  CURRENT_DATE = Time.new.year.to_s + "-" + Time.new.month.to_s + "-" + Time.new.day.to_s
```
ben-davis-devops, line 6: clever use of a constant

lib/messages.rb
```diff
@@ -0,0 +1,37 @@
...
+  end  
+
+  def self.create(message:)
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else 
+      connection = PG.connect(dbname: 'chitter')
+    end
+
+     result = connection.exec("INSERT INTO peeps (message, date) VALUES('#{message}', '#{CURRENT_DATE}') RETURNING id, message, date;")
```
ben-davis-devops, line 34: could alter this slightly to consider protecting against SQL injection:
"INSERT INTO peeps (message, date) VALUES($1, $2) RETURNING id, message, date;", [message, date])

PR #94, opened 2021-08-16
Hamid

p4r7hum4n:


siellsiell (commented):
Hey Hamid,

thanks for submitting your solution! You've demonstrated a good grasp of working with databases and the MVC pattern.

For the future, I'd encourage you to pay extra attention to the quality and strength of your tests, and to keep it high especially as you add new features. In some cases your tests don't test everything that they should be testing, e.g. the new `date` column you added. I left some comments with more details on that. 

Keep up the good work!

PR #101, opened 2021-10-15
Daniel Szabo Week6 Individual challenge

DannyL90:


app.rb
```diff
@@ -1,9 +1,22 @@
-require 'sinatra/base'
+require "sinatra/base"
+require "./lib/peep"
 
 class Chitter < Sinatra::Base
-  get '/test' do
-    'Test page'
+
+  get "/" do
```
delexii, line 6: I would probably suggest using a different view for this reverse order of messages. You could try and call it peeps/all/reversed

PR #102, opened 2021-10-15
chitter challenge

pchan2:


spec/features/viewing_peeps_spec.rb
```diff
@@ -0,0 +1,20 @@
+feature 'See all messages in a browser' do
+  scenario 'To see what people are doing, I want to see all the messages' do
+    Peep.add(name: 'Patrick', message: 'message 1', date: '2021-10-17')
+    Peep.add(name: 'Jon', message: 'message 2', date: '2021-10-18')
+    Peep.add(name: 'Mary', message: 'message 3', date: '2021-10-19')
+
```
EdwardAndress, line 6: This is really clear and easy to read but how can you be sure that, for example, `Patrick` sent the `message 1` peep?

HINT: rather than placing an assertion on what the page contains, you can grab individual elements and make assertions about what they contain.

PR #103, opened 2021-10-15
Matthew Read Chitter Challenge 15/10/21

matthewreadbass:


feedXia (commented):
Hey Matthew!

Your code was clean and easy to read. I really enjoyed seeing how you do things & all the references! 

The way you structured your era files was really interesting.

I learned a lot!

db/migrations/02_add_time_column.sql
```diff
@@ -0,0 +1,4 @@
+In psql bookmark_manager & bookmark_manager_test databses
```
EdwardAndress, line 1: very nice

app.rb
```diff
@@ -1,8 +1,30 @@
-require 'sinatra/base'
+require "sinatra"
+require "sinatra/base"
+require "./lib/peep"
```
EdwardAndress, line 3: This looks great on the whole and clearly demonstrates some of the week's learning objectvives. The one thing I would recommend is that you revisit RESTful routing.

lib/peep.rb
```diff
@@ -0,0 +1,46 @@
+require "pg"
+require "time"
+
```
EdwardAndress, line 3: Did you consider doing something like this, to build hashes with all the details?

```ruby
peeps.map do |peep|
   { :message => peep["message"] , date: peep['date'], time: peep['time']}
end
```

lib/peep.rb
```diff
@@ -0,0 +1,46 @@
...
+      connection = PG.connect(dbname: "chitter_test")
+    else
+      connection = PG.connect(dbname: "chitter")
+    end
+    peeps = connection.exec("SELECT * FROM peeps")
+    peeps.map do |peep|
+      { :time => peep["time"] }
+    end
+  end
+end
```
EdwardAndress, line 46: You've done a good job of demonstrating some key learning objectives here (writing to and reading from a database).  There is some repetition, however, so the next step would probably be to remove that (rather than add more features).  How could you go about doing that and why is it important?

PR #104, opened 2021-10-15
Chitter Challenge - Lauren

adashtra:


lib/peep.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def self.post(message:)
+    if ENV["ENVIRONMENT"] == "test"
+      connection = PG.connect(dbname: "chitter_test")
+    else
+      connection = PG.connect(dbname: "chitter")
+    end
+
+    connection.exec("INSERT INTO peeps (message) VALUES('#{message}')")
```
FrancineBlanc, line 24: You could DRY up your code here by adding the if/else statement in a separate method, and then calling that in `self.all` and `self.post`

spec/features/viewing_peeps_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  scenario "Peeps are displayed in reverse chronological order" do
+    Peep.post(message: "This is the oldest peep!")
+    Peep.post(message: "This is the most recent peep!")
+
+    visit("/peeps")
+    expect(page).to have_content("This is the most recent peep! - Posted on 15/10/2021 This is the oldest peep! - Posted on 15/10/2021")
+  end
+end
```
FrancineBlanc, line 21: Clear, easy to read tests.

lib/peep.rb
FrancineBlanc, line 24: You may want to think about SQL injection and protecting your database against malicious attacks.

app.rb
```diff
@@ -1,8 +1,27 @@
...
-  get '/test' do
-    'Test page'
+  get "/test" do
+    "Test page"
+  end
+
+  get "/" do
+    "Chitter"
+  end
+
```
FrancineBlanc, line 12: You may want to include this in a separate erb file, so that this route isn't handling any "view" data.

EdwardAndress:
Hey Lauren! Nice work. I'd say that you've demonstrated the key learning objectives here – well done! Thinking ahead from here – what would you plan for the next user story have been?

PR #112, opened 2021-10-15
Chitter Challenge

SamJT2021:


lib/database_connection.rb
```diff
@@ -0,0 +1,15 @@
+require 'pg'
+
+class DatabaseConnection
+  def self.setup(dbname)
+    @connection = PG.connect(dbname: dbname)
+  end
+
+  class << self
+    attr_reader :connection
```
lewiscj97, line 9: nice

spec/features/creating_peeps_spec.rb
```diff
@@ -0,0 +1,10 @@
+feature 'Create peep' do
```
lewiscj97, line 1: good test

spec/setup_test_database.rb
```diff
@@ -5,7 +5,8 @@ def setup_test_database
   connection.exec("TRUNCATE peeps;")
 end
 
-def add_row_to_test_database
+def add_rows_to_test_database
```
lewiscj97, line 8: this is a smart little helper method

views/create_peep.erb
```diff
@@ -0,0 +1,9 @@
+<div class="centered-content">
+  <form action="/new-peep" method="post">
+    <h2 for="message" class="create-peep-header">Message:</h2>
+    <input type="text" name="message" id="message" placeholder="What's on your mind?" maxlength="60"required>
```
lewiscj97, line 4: Does maxlength say that the user can only put in 60 characters? If so, that's really smart

lewiscj97:
Nice job, it looks and works really nicely. Left a few minor comments, but these are mostly just personal preference. I cloned the repo and had a play around, it works great with the `required` fields for the message and username and the `maxlength` tag for inputs is something I have not seen before - makes the website really nice to operate. Nice one!

app.rb
```diff
@@ -1,8 +1,29 @@
...
+    @peeps = Peep.all
+    erb(:peeps)
+  end
+
+  get '/create-peep' do
+    erb(:create_peep)
+  end
+
+  post '/new-peep' do
+    Peep.create(params[:message], params[:username], Date.today.strftime('%d/%m/%Y'))
```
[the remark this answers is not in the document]
Katerina-codes: I would suggest you look at RESTful routes. See below for more details:
Really good work on this challenge but I would suggest you look at RESTful routes:

PR #113, opened 2021-10-15
Lewis - Chitter Challenge

lewiscj97:


app.rb
```diff
@@ -1,8 +1,40 @@
...
+
+  post '/search' do
+    formatted_query = params[:query].gsub(" ", "_")
+    redirect("search/#{formatted_query}")
+  end
+
+  get '/search/:query' do
+    @query = params[:query].gsub("_", " ")
+    @peeps = Peep.filter(@query)
+    erb(:result)
```
troysnowden, line 37: Nice use of a route specifically for the filter aspect, and using the URL param to pass the query across.

lib/database_connection.rb
```diff
@@ -0,0 +1,11 @@
...
+
+class DatabaseConnection
+  def self.setup(dbname)
+    @connection = PG.connect(dbname: dbname)
+  end
+
+  def self.query(sql, params = [])
+    @connection.exec_params(sql, params)
+  end
+end
```
troysnowden, line 11: I like extracting the database connection into its own class, following your Single Responsibility Principle

lib/date_formatter.rb
```diff
@@ -0,0 +1,11 @@
...
+
+class DateFormatter
+  STRP_TIME_FORMAT = '%Y-%m-%d'.freeze
+  STRF_TIME_FORMAT = '%d/%m/%y'.freeze
+
+  def self.format_date(date)
+    date = Date.strptime(date, STRP_TIME_FORMAT)
+    date.strftime(STRF_TIME_FORMAT)
+  end
+end
```
troysnowden, line 11: Again good extraction of logic into its own class

lib/peep.rb
```diff
@@ -0,0 +1,48 @@
...
+class Peep
+  FILTER_QUERY = 
+  "SELECT * FROM peeps WHERE UPPER(message) LIKE $1
+  ORDER BY date DESC, id DESC;".freeze
+  INSERT_NO_DATE_QUERY = 
+  "INSERT INTO peeps(message) VALUES($1) RETURNING id, message, date".freeze
+  INSERT_QUERY = 
+  "INSERT INTO peeps(message, date) VALUES($1, $2) RETURNING id, message, date".freeze
+  ALL_PEEPS_QUERY =
+  "SELECT * FROM peeps ORDER BY date DESC, id DESC;".freeze
```
troysnowden, line 13: I question the need for these class constants, given they are only used once each? Makes the query a little neater however I agree.

lib/peep_factory.rb
```diff
@@ -0,0 +1,8 @@
+require_relative 'peep'
+
+# This factory takes a SQL response and wraps it in an instance of the peep class
+class PeepFactory
+  def self.wrap_peep_response_in_object(peep)
+    Peep.new(id: peep['id'], message: peep['message'], date: peep['date'])
+  end
+end
```
troysnowden, line 8: Nice use of factory pattern

spec/features/chitter_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+    DatabaseConnection.query("INSERT INTO peeps(message) VALUES('This is a newer peep');")
+    
+    visit('/')
+    click_button 'Search'
+
+    fill_in 'query', with: 'peep'
+    click_button 'Search'
+
+    expect(page).to have_content 'This is a newer peep'
+    expect(page).to have_content 'This is an older peep'
```
troysnowden, line 52: Nice test for the filter feature

views/index.erb
```diff
@@ -0,0 +1,23 @@
+<div class="new_post">
+  <form method="post" action="/new">
+    <textarea class="text-input" name="message" placeholder="Enter a peep (max 200 char)..."></textarea><br>
```
troysnowden, line 3: 'textarea' tag makes much more sense than a regular input...oops.

views/index.erb
```diff
@@ -0,0 +1,23 @@
...
+    <input class="submit button" type="submit" value="Post">
+  </form>
+
+  <%= flash[:empty] %>
```
troysnowden, line 7: So I'm guessing this flash doesn't come up if the empty key of the flash param has not been populated.

views/index.erb
```diff
@@ -0,0 +1,23 @@
...
+</div>
+
+<div class="search">
+  <form method="get" action="/search">
+    <input class="search button" type="submit" value="Search">
+  </form>
+</div>
```
troysnowden, line 14: Would it make more sense from a user's perspective to have the search bar on the index page that contains the peeps?

lib/peep.rb
```diff
@@ -0,0 +1,48 @@
...
+        INSERT_NO_DATE_QUERY, [message])
+    else
+      response = DatabaseConnection.query(
+        INSERT_QUERY, [message, date])
+    end
+    PeepFactory.wrap_peep_response_in_object(response.first)
+  end
+
+  def self.filter(message)
+    # upcase to match message case-insensitive
```
Jestfer, line 40: I like comments that explain the **WHY** and not the **WHAT**. Keep it up 👍

Jestfer:
Great work on documentation for the challenge! I love the separation of concerns into different markdown files. Keep it up!

An excellent solution! You have demonstrated the learning objectives for this unit by implementing all of the user stories!

Well done again, really good submission!

PR #118, opened 2021-10-16
Makers Chitter challenge

2Rads:


lib/database.rb
```diff
@@ -0,0 +1,24 @@
...
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
+    return connection
+  end
+
+  def self.get_messages
+    connection = Database.connect
+    result = connection.exec("SELECT * FROM peeps ORDER BY date, time;")
```
pchan2, line 16: Would this be `ORDER BY date, time DESC` to have the most recent messages appear first?

lib/database.rb
```diff
@@ -0,0 +1,24 @@
...
+    result
+  end
+
+  def self.add_message(text:)
+    connection = Database.connect
+    connection.exec("INSERT INTO peeps (message, date, time) VALUES('#{text}','#{Date.today}','#{Time.now}')")
```
pchan2, line 22: For securing the app to avoid SQL injection via input field, use `exec_params.`

```
connection.exec_params(
  "INSERT INTO peeps (message, date, time) VALUES($1, $2, $3) RETURNING id, message, date, time;", [message, date, time]
)
```

views/login.erb
```diff
@@ -0,0 +1,7 @@
+<form action='/login' method='post'>
+  <input type="text" name="user" placeholder="username" required/>
+  <input type="password" name="password" placeholder="password" required/>
+  <input type="submit" value="Submit" />
+</form>
+
+<p>New? <a href='login/new'>create an account</a></p>
```
pchan2, line 7: **C**reate an account 👍

lib/users.rb
```diff
@@ -0,0 +1,17 @@
+class Users
+  def self.create(username:, password:)
+    connection = Database.connect
+    result = connection.exec("SELECT username FROM users WHERE username = '#{username}'")
+    if result.ntuples != 0
+      return false
+    end
+    connection.exec("INSERT INTO users (username, password) VALUES('#{username}','#{password}')")
```
pchan2, line 8: For securing the app to avoid SQL injection via input field, use exec_params.

connection.exec_params(
  "INSERT INTO users (username, password) VALUES($1, $2) RETURNING id, username, password;", [sername, password]
)

PR #122, opened 2021-10-17
Chitter challenge

AishaDorsett:
Currently incomplete as I haven't added a filter for the peeps

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -7,16 +7,16 @@ class Chitter < Sinatra::Base
   end
 
   get '/chitter' do
-    @peeps = Chitter_Page.view
+    @peeps = ChitterPage.view
     erb :peeps
   end
 
   get '/chitter/new' do
-    erb :"new_message"
+    erb :new_message
   end
 
   post '/chitter/create' do
-    Chitter_Page.create(params[:message])
+    ChitterPage.create(params[:message])
     redirect "/chitter"
   end
 
--- a/lib/chitter_homepage.rb
+++ b/lib/chitter_homepage.rb
@@ -1,6 +1,6 @@
 require 'pg'
 
-class Chitter_Page
+class ChitterPage
   def self.view
     if ENV["ENVIRONMENT"] == "test"
       connection = PG.connect(dbname: "chitter_test")
@@ -14,7 +14,6 @@ class Chitter_Page
     end
   end
 
-
   def self.create(message)
     if ENV["ENVIRONMENT"] == "test"
       connection = PG.connect(dbname: "chitter_test")
@@ -24,5 +23,4 @@ class Chitter_Page
 
     connection.exec("INSERT INTO peeps (message) VALUES('#{message}')")
   end
-
 end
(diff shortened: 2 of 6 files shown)


app.rb
```diff
@@ -1,9 +1,24 @@
 require 'sinatra/base'
+require './lib/chitter_homepage'
 
 class Chitter < Sinatra::Base
   get '/test' do
     'Test page'
   end
```
pinteck, line 7: You can remove the test page

db/migrations/01_create_chitter_table.sql
```diff
@@ -1 +1,11 @@
...
+
+/* add a date column*/
+ALTER TABLE peeps
+ADD COLUMN posted_at TIMESTAMP;
+
+ALTER TABLE peeps
+ALTER COLUMN date_posted SET DEFAULT now();
+
+/* Order peeps by descending order */
+SELECT * FROM peeps ORDER BY date_posted DESC;
```
pinteck, line 11: Great practice that you document the sql commands.

app.rb
```diff
@@ -1,9 +1,24 @@
...
+  get '/chitter' do
+    @peeps = ChitterPage.view
+    erb :peeps
+  end
+
+  get '/chitter/new' do
+    erb :new_message
+  end
+
+  post '/chitter/create' do
```
pinteck, line 18: When you use 'post' you can use the same route name, this helps to have a restful api.

lib/chitter_homepage.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def self.create(message)
+    if ENV["ENVIRONMENT"] == "test"
+      connection = PG.connect(dbname: "chitter_test")
+    else
+      connection = PG.connect(dbname: "chitter")
+    end 
+
+    connection.exec("INSERT INTO peeps (message) VALUES('#{message}')")
```
pinteck, line 24: In order to protect the app from injection, you can use `.exec_params` method instead of `.exec`.
And you can pass the message with this syntax:
`connection.exec_params("INSERT INTO peeps (message) VALUES($1);", [message])`

lib/chitter_homepage.rb
```diff
@@ -0,0 +1,26 @@
+require 'pg'
+
+class ChitterPage
+  def self.view
+    if ENV["ENVIRONMENT"] == "test"
+      connection = PG.connect(dbname: "chitter_test")
+    else
+      connection = PG.connect(dbname: "chitter")
+    end
+    total = connection.exec("SELECT * FROM peeps ORDER BY id DESC;")
```
pinteck, line 10: You can also use the `SELECT * FROM peeps ORDER BY date_posted DESC;` command here to call the peeps by descending order directly from the database.

views/peeps.erb
```diff
@@ -0,0 +1,33 @@
+<style>
```
pinteck, line 1: Appreciate that you took time to style your app. You can take a screenshot of the page and add it to your readme to give a quick sneak peek.

PR #124, opened 2021-10-17
Chitter Challenge

pinteck:


spec/features/add_a_new_peep_spec.rb
```diff
@@ -0,0 +1,9 @@
+feature 'Add' do
```
AishaDorsett, line 1: its great that you've separated the features tests across multiple files and for each feature. It makes it so much easier to read/see what you have done especially as this challenge has many components.

spec/features/homepage_spec.rb
```diff
@@ -0,0 +1,6 @@
+feature 'Homepage' do
+  scenario 'has a title on the homepage' do
+    visit('/')
```
AishaDorsett, line 3: you could make the chitter homepage route '/chitter' and make the messages route '/chitter/messages' as I think Eddie said it makes more sense for the user to visit those routes

spec/features/show_all_messages_spec.rb
```diff
@@ -0,0 +1,14 @@
+feature 'Messages' do
+  scenario 'show all messages' do
+    connection = PG.connect(dbname: 'chitter_test')
+    
+    connection.exec("INSERT INTO peeps (message) VALUES ('Today''s first peep');")
+    connection.exec("INSERT INTO peeps (message) VALUES ('Today''s second peep');")
+    connection.exec("INSERT INTO peeps (message) VALUES ('Today''s third peep');")
+
+    visit '/messages'
+    expect(page).to have_content "Today's first peep"
```
AishaDorsett, line 10: very clean code

spec/message_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      #add the test data
+      connection.exec("INSERT INTO peeps (message) VALUES ('Today''s first peep');")
+      connection.exec("INSERT INTO peeps (message) VALUES ('Today''s second peep');")
+      connection.exec("INSERT INTO peeps (message) VALUES ('Today''s third peep');")
+
+      messages = Message.all 
+      list_of_messages = messages.map { |item| item.message }
+
+      expect(list_of_messages).to eq ["Today's first peep","Today's second peep","Today's third peep"]
```
AishaDorsett, line 16: this is a great way of testing the messages. I wasn't sure how to do it properly in a unit test but this makes a lot of sense.

views/find.erb
```diff
@@ -0,0 +1,14 @@
+<ul>
```
AishaDorsett, line 1: again, I really like that you've separated the views across multiple files for each feature and added HTML styling to it.

views/messages.erb
```diff
@@ -0,0 +1,30 @@
+ <div style="text-align:center;">   
+    <h1 style="color:#5cd65c;font-family:courier;font-size:350%;"> Chitter </h1>
+
+    <form action="/messages" method="post">
+        <input style="font-family:courier;padding:13px 80px;text-align:center;font-size:16" type="text" name="new_message" placeholder="New Peep" maxlength="60" />
```
AishaDorsett, line 5: instead of using inline styling, you could add a <style> tag at the top of your documents and add all of your styling in there to keep your code clean/readable, whilst maintaining the styling elements you have added - without using CSS :) 
(if you are using multiple h1 tags and want to style them each differently for example, you could achieve this by adding an "id" to each h1 tag and then reference that id within the <style></style> tag - not sure if this makes sense).

PR #125, opened 2021-10-17
Chitter Challenge 

FrancineBlanc:
This is my submission for the Chitter Challenge, which satisfies all of the user stories. 
For next steps I would like to incorporate:
-  validation, so that users need to enter a username and password to use the web app;
- comments so that users can comment on each other's peeps.


lib/peep.rb
```diff
@@ -0,0 +1,33 @@
...
+  def self.sort_by_date
+    result = DatabaseConnection.query("SELECT id, username, message, post_date, TO_CHAR(post_date, 'dd/mm/yyyy') FROM peeps ORDER BY post_date DESC;")
+    result.map do |row|
+      { id: row["id"], username: row["username"], message: row["message"], post_date: row["to_char"] }
+    end
+  end
+
+  def self.search(search_term)
+    result = DatabaseConnection.query("SELECT id, username, message, post_date, TO_CHAR(post_date, 'dd/mm/yyyy') FROM peeps WHERE message LIKE '%#{search_term}%';")
+    @@search_results = result.map do |row|
```
adashtra, line 25: Just wondering - what does the @@ in search_results do?

PR #127, opened 2021-10-18
chitter challenge

ewa-j:


lib/peep.rb
```diff
@@ -0,0 +1,47 @@
...
+  attr_reader :id, :message, :date_posted, :keyword
+
+  def initialize(id:, message:, date_posted:)
+    @id = id
+    @message = message
+    @date_posted = date_posted
+  end
+
+  def self.all
+    if ENV["ENVIRONMENT"] == "test"
```
dannehdan, line 13: You could use a one-liner here to improve your coverage and reduce the lines-per-method in your peep.rb file. You could also separate this into a method to DRY up your code (you can see what I mean in my code)

app.rb
```diff
@@ -1,8 +1,38 @@
...
+require "./lib/peep"
 
 class Chitter < Sinatra::Base
-  get '/test' do
-    'Test page'
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  get "/test" do
```
dannehdan, line 10: You could remove your test page when you're finished to clean up this file if it's not going to be used again

app.rb
```diff
@@ -1,8 +1,38 @@
...
+  get "/" do
+    @peeps = Peep.all
+    erb(:index)
+  end
+
+  get "/new" do
+    erb(:new)
+  end
+
+  post "/" do
```
dannehdan, line 23: I'm not sure if having two routes with the '/' path is confusing? Maybe this could be called '/post-peep' or something that then redirects to '/'

views/search.erb
```diff
@@ -0,0 +1,4 @@
+<form action="/filtered" method="post">
```
dannehdan, line 1: This would be really cool if this page was on your homepage, so the search function is available from the homescreen? Really cool that you got it to work though!

views/new.erb
```diff
@@ -0,0 +1,4 @@
+<form action="/" method="post">
```
dannehdan, line 1: How do you get to this page? Do you have to type '/new' in the url? If so, maybe a button on the homepage could take you here instead?

lib/peep.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def self.filter(keyword:)
+    if ENV["ENVIRONMENT"] == "test"
+      connection = PG.connect(dbname: "chitter_test")
+    else
+      connection = PG.connect(dbname: "chitter")
+    end
+
+    result = connection.exec("SELECT * FROM peeps WHERE message ~* '#{keyword}' ORDER BY date_posted DESC;")
```
dannehdan, line 42: Super impressed by your search function - looks really clean!

lib/peep.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def self.create(message:)
+    if ENV["ENVIRONMENT"] == "test"
+      connection = PG.connect(dbname: "chitter_test")
+    else
+      connection = PG.connect(dbname: "chitter")
+    end
+
+    connection.exec("INSERT INTO peeps (message) VALUES('#{message}')")
```
dannehdan, line 32: You should look up the SQL Injection walkthrough from the bookmark manager challenge (https://github.com/makersacademy/course/blob/main/apprenticeships_bookmark_manager/walkthroughs/12.md) to make sure that someone can't break your program

With this style of query you've written, someone could write a peep that says 'DROP TABLE' and delete your database!

dannehdan (commented):
Really good job! I'm so impressed with the fact you got filters to work with your program - hopefully the rest of my review makes sense, but feel free to ask me if you aren't sure!

PR #129, opened 2021-10-18
Steve J - Chitter task

stevej763:


app.rb
```diff
@@ -1,8 +1,69 @@
 require 'sinatra/base'
+require './lib/services/peep_service'
+require './lib/repositories/peep_repository'
+require './lib/database/database_connection_setup'
 
 class Chitter < Sinatra::Base
-  get '/test' do
-    'Test page'
+
```
oceansize, line 7: Thank you for deleting this bloody route!

app.rb
```diff
@@ -1,8 +1,69 @@
...
+    redirect('/peeps')
+  end
+
+  private 
+
+  def get_peeps(filter)
+    if @peep_order == ORDER_CHRONOLOGICAL
+      @peeps = @peep_service.find_all_peeps
+    else
+      get_peeps_reverce_chronological_order(filter)
```
oceansize, line 47: `reverce`? Is that Italian?

lib/repositories/peep_repository.rb
```diff
@@ -0,0 +1,29 @@
+require_relative '../database/database_connection'
+require_relative '../domain/peep'
+
+class PeepRepository
+  def initialize
+    
+  end
```
oceansize, line 7: No need to declare `initialize` unless you're actually doing something with it.

spec/setup_test_database.rb
```diff
@@ -1,11 +1,38 @@
...
+end
+
+def add_peep_for_filter
+  connection = connect
+  connection.exec("INSERT INTO peeps (message, time_stamp) values ('custom peep', '#{DateTime.new(2021, 10, 14)}');")
+end
+
+private
+
+def connect
```
oceansize, line 36: Lovely!

PR #130, opened 2021-10-18
Xia Chitter Challenge

feedXia:


app.rb
```diff
@@ -1,8 +1,19 @@
...
-    'Test page'
+  # configure :development do
+  #   register Sinatra::Reloader
+  # end
+
+  get "/test" do
+    "Test page"
+  end
+
+  get "/peeps" do
```
matthewreadbass, line 14: Nice and concise - easy to follow the train of thought and the route that the code will follow

PR #131, opened 2021-10-18
Chitter-Challenge-Stella

stellamartin222:


lib/database_connection.rb
```diff
@@ -0,0 +1,15 @@
...
+      end
+
+    def self.connection
+        @connection
+    end
+
+    def self.query(sql, params = [])
+        @connection.exec_params(sql, params)
+    end
+end
```
thomas99baxter, line 15: 🤌🏼

lib/database_connection_setup.rb
```diff
@@ -0,0 +1,9 @@
+require './lib/database_connection'
+
+if ENV['ENVIRONMENT'] == 'test'
+  DatabaseConnection.setup('chitter_test')
+else
+  # :nocov:
+  DatabaseConnection.setup('chitter')
+  # :nocov:
+end
```
thomas99baxter, line 9: just personal preference, but i like to use this in a ternary. do think this is really clean though!

spec/web_helpers.rb
```diff
@@ -0,0 +1,9 @@
+require 'pg'
+
+def test_db_inserts
+    connection = PG.connect(dbname: 'chitter_test')
+      # Add test data
+      connection.exec("INSERT INTO peeps (message) VALUES ('I hate mondays');")
+      connection.exec("INSERT INTO peeps (message) VALUES ('Lasagne is my favourite');")
+      connection.exec("INSERT INTO peeps (message) VALUES ('Garfield is a cool guy');")
```
thomas99baxter, line 8: love

views/messages.erb
```diff
@@ -0,0 +1,12 @@
...
+    <input type='text' name='post_message' class='post_input_box'>
+    <input value='post' type='submit' class='post_button'>
+</form>
+
+<% @messages.each_with_index do | message, index |%>
+    <section id=<%=index%>
+        <p> <%=message.date%> </p>
+        <p> <%=message.message%> </p>
+    </section>
+<% end %>
```
thomas99baxter, line 12: would be good to have this wrapped in <body> tags. think its done automatically in some browsers but good to have it in for compatibility

views/layout.erb
```diff
@@ -0,0 +1,10 @@
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Bookmark Manager</title>
+    <link rel="stylesheet" href="<%='app.css'%>">
+  </head>
+  <h1>CHITTER</h1>
+  <a href="/messages">Home</a>
+  <%= yield %>
```
thomas99baxter, line 9: yield?

views/layout.erb
```diff
@@ -0,0 +1,10 @@
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Bookmark Manager</title>
+    <link rel="stylesheet" href="<%='app.css'%>">
+  </head>
+  <h1>CHITTER</h1>
+  <a href="/messages">Home</a>
```
thomas99baxter, line 8: body tags would be good here

app.rb
```diff
@@ -1,9 +1,25 @@
...
+  get '/' do
+    erb(:index)
+  end
+
+  get '/messages' do
+    @messages = Message.all
+    erb(:messages)
+  end
+
+  post '/post-message' do
```
Katerina-codes, line 19: Overall you've done a good attempt with this challenge. 

I would suggest having a look at RESTful routes. Here your post would changes to **post '/messages' do**


See below for more examples:

PR #133, opened 2021-10-18
tom baxter

thomas99baxter:


app.rb
```diff
@@ -1,8 +1,37 @@
...
+    $filtered_peeps = PeepAccessor.filter_peeps_by_keyword(params[:keyword])
+    redirect '/display_filtered_peeps'
+  end
+
+  get '/display_filtered_peeps' do
+    @filtered_peeps = $filtered_peeps
+    erb(:display_filtered_peeps)
   end
 
   run! if app_file == $0
```
stellamartin222, line 37: Very clear and easy to read naming conventions for endpoints

lib/peep_accessor.rb
```diff
@@ -0,0 +1,32 @@
...
+    response = db_connection.exec "SELECT * FROM peeps WHERE message LIKE '%#{keyword}%'"
+
+    response.each do |row| 
+      filtered_peeps << row["message"]
+    end
+
+    return filtered_peeps
+  end
+
+end
```
stellamartin222, line 32: I like that you have extracted the connection from this class completely to keep it DRY

lib/peep_accessor.rb
```diff
@@ -0,0 +1,32 @@
...
+require_relative 'helpers/db_connector'
+class PeepAccessor
+  def initialize; end
+
+  def self.all
+    peeps_arr = []
+  
+    rs = db_connection.exec 'SELECT * FROM peeps'
+
+    rs.each do |row|
```
stellamartin222, line 11: could this be done as a map function?

spec/peep_accessor_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    it "should return an empty array when it doesnt match" do
+      described_class.add_peep('this is another peep!')
+      described_class.add_peep('this is a peep that mentions Borat')
+      expect(described_class.filter_peeps_by_keyword('Bruno')).to eq(
+        []
+      )
+    end
+  end
+end
```
stellamartin222, line 51: Excellent testing library!

views/add_peep.erb
```diff
@@ -0,0 +1,4 @@
+<form method="post" action="/add_peep">
+  <input type="text" name="peep" placeholder="your peep here"required>
+  <input type="submit" name="Submit">
+</form>
```
stellamartin222, line 4: I like the simplicity of the form, it is very user friendly

views/index.erb
```diff
@@ -0,0 +1,18 @@
+<h1>Chitter</h1>
```
stellamartin222, line 1: If you wanted this to be displayed on every page an easy way would be to extract it into a layout erb and place the h1 in the header

views/index.erb
```diff
@@ -0,0 +1,18 @@
...
+  <form method="GET" action="/add_peep">
+    <input type="submit" value="Add Peep">
+  </form>
+</div>
+
+<div id="filter-the-peeps"> 
+  <form method="GET" action="/filter_peeps">
+    <input type="submit" value="Filter Peeps">
+  </form>
+</div>
```
stellamartin222, line 18: I can see you were in the process of extracting these two forms out into their own separate erbs which would allow each erb to follow the single responsibility principle

stellamartin222 (commented):
Great chitter challenge, you show clear understanding of backend and front end connections. You have done so much for just a single days work, massive well done

views/add_peep.erb
thomas99baxter: thanks!

lib/peep_accessor.rb
```diff
@@ -0,0 +1,32 @@
+require_relative 'peep'
+require_relative 'helpers/db_connector'
+class PeepAccessor
+  def initialize; end
```
oceansize, line 4: You don't need to declare `initialize` if you're not extending the method — Ruby already does this for you in the background. We do this only if we want the class initialisation to create some state, spin up another object or run another kind of operation.

oceansize: Also if you delete this, your app will start behaving much more as expected.

lib/peep_accessor.rb
oceansize: Agreed, it's _noice_

app.rb
```diff
@@ -1,8 +1,37 @@
...
+    PeepAccessor.add_peep(params[:peep])
+    redirect '/'
+  end
+
+  get '/filter_peeps' do
+    erb(:filter_peeps)
+  end
+
+  post "/filter_peeps" do
+    # i will go to hell for doing this
```
oceansize, line 27: https://media.giphy.com/media/xT9KVg8gkDEyJIrVdK/giphy.gif

lib/peep_accessor.rb
thomas99baxter: I'll have to have a _peep_ at this once we've got the final project finished!

thomas99baxter: I'll have to have a _peep_ at this once we've got the final project finished!

PR #134, opened 2021-10-18
for code review - Oisin

duckduckboom:


README-Oisin.md
```diff
@@ -0,0 +1,28 @@
...
+I want to post a message (peep) to chitter -->
+
+Who? - Maker
+Class - peep
+
+What is needed to achieve this? 
+A page to submit a peep
+A form to save a peep to the database
+
+
```
michelle-watkins, line 28: What do you think about adding a section for cloning/running instructions? You could include how to clone your code from GitHub, how to run your app (`rackup`) and how to create the database locally on their machine.

app.rb
```diff
@@ -5,5 +5,13 @@ class Chitter < Sinatra::Base
     'Test page'
   end
 
+  get '/chitter' do
+    erb :chitter
+  end
```
michelle-watkins, line 10: What do you think about replacing '/chitter' with '/'? This would mean that the user would directly see your 'Chitter feed' title without having to explicitly put that route into the URL.

spec/features/chitter_spec.rb
```diff
@@ -0,0 +1,16 @@
+feature 'Viewing chitter' do
+    scenario 'visiting the chitter page' do
+        visit('/chitter')
+        expect(page).to have_content("Chitter feed")
+    end
+end
```
michelle-watkins, line 6: This test is really clear and easy to understand.

spec/features/chitter_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+
+feature 'Posting a peep' do
+    scenario 'visiting the page' do
+        visit('post-a-peep')
```
michelle-watkins, line 10: Does this work without the '/' before the page name (post-a-peep)? I've just tried running `rspec` to check and your tests pass. So maybe '/' is not always needed. Interesting to know!

spec/features/chitter_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+        expect(page).to have_content("Post a Peep")
+        expect(page).to have_content("Your peep:")
+        expect(page).to have_field("peep")
+        expect(page).to have_button("Submit")
```
michelle-watkins, line 14: I like how you've used different Capybara commands.

views/post_a_peep.erb
```diff
@@ -0,0 +1,8 @@
+Post a Peep
+<form action="/name" method="post">
+
+    <label for="name">Your peep:</label><br>
+    <input type="text" id="peep" name="peep"><br>
+    <input type="submit" value="Submit">
+
+</form>
```
michelle-watkins, line 8: I really like that you thought of using a form. It looks clear in the browser that a user can 'Submit' a message. I didn't think of doing this!

views/post_a_peep.erb
```diff
@@ -0,0 +1,8 @@
+Post a Peep
+<form action="/name" method="post">
+
+    <label for="name">Your peep:</label><br>
+    <input type="text" id="peep" name="peep"><br>
+    <input type="submit" value="Submit">
```
michelle-watkins, line 6: I noticed when I clicked 'Submit' that it currently goes to '/name' but I can't see it referenced in your code. Could you explain what would have happened on the next page?

app.rb
Katerina-codes: Good job with this challenge. I would suggest you look at RESTful routes. See below for more details:

PR #136, opened 2021-10-18
allowing retrieval of peeps

sunnydonaldson:


app.rb
```diff
@@ -1,9 +1,16 @@
 require 'sinatra/base'
+require './lib/Peep.rb'
 
 class Chitter < Sinatra::Base
   get '/test' do
     'Test page'
   end
 
+  get '/' do
+    @peeps = Peep.all_peeps()
```
EdwardAndress, line 10: In Ruby, if there's no params, you don't need the brackets :)

spec/spec_helper.rb
```diff
@@ -41,6 +41,7 @@
 RSpec.configure do |config|
   config.before(:each) do
     setup_test_database
+    add_row_to_test_database
```
EdwardAndress, line 44: Well done getting this to work.  If you imagine a new developer trying to get to grips with this codebase though, would it be clearer to call `add_row_to_test_database` in the spec file?

spec/units/peep_spec.rb
```diff
@@ -0,0 +1,9 @@
+require "peep"
+
+describe Peep do
+  describe ".all_peeps" do
+    it "returns an array of all the peeps" do
```
EdwardAndress, line 5: Here you're stating that the method should return an array of peeps but then you only actually test for the message of the first peep.

How could you improve this test?

PR #137, opened 2021-10-20
Joe's work

Joseph112358:
Apologies for the late submission ( I was ill on Monday)


lib/peeps.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def initialize(id: ,message:, name:, date:)
+    @id = id
+    @message = message
+    @name = name
+    @date = date
+    
+  end
+
+  def self.list_peeps()
```
EdwardAndress, line 16: In ruby, if there are no args, you can leave out the brackets

lib/peeps.rb
```diff
@@ -0,0 +1,42 @@
...
+      Peeps.new(id: peep['id'], message: peep['message'], name: peep['name'], date: peep['date'])
+    end
+  end
+
+  def self.add_peep(message: , name:, date:)
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
```
EdwardAndress, line 33: What's the intention around assigning these instance vars?  Are they used somewhere?

PR #138, opened 2021-11-05
Patrick Barsch - Chitter Challenge

PatrickBarsch:
User story 1 - the journey begins

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -5,5 +5,9 @@ class Chitter < Sinatra::Base
     'Test page'
   end
 
+  get '/messages' do
+    'This is a peep!'
+  end
+
   run! if app_file == $0
 end
--- a/spec/features/view_messages_spec.rb
+++ b/spec/features/view_messages_spec.rb
@@ -1,0 +1,7 @@
+feature 'Viewing messages' do
+  scenario 'A user can see messages' do
+    add_row_to_test_database
+    visit '/messages'
+    expect(page).to have_content('This is a peep!') 
+  end
+end


app.rb
```diff
@@ -1,8 +1,19 @@
 require 'sinatra/base'
+require './lib/peep'
+require './database_connection_setup'
 
 class Chitter < Sinatra::Base
-  get '/test' do
-    'Test page'
+
+  get '/messages' do
+    Peep.toggle_order if params['sorting']
```
EdwardAndress, line 8: This is a clever solution but I'm wondering if it's a little dangerous that the order can be toggled and then left in the toggled state.  What if you were to use Peep.all in another context?

EdwardAndress: An alternative would be to have an optional second parameter on Peep.all which could specify the ordering.  Or, another one, would be to just always return the Peeps in reverse chronological order (I don't think the requirements ask for any flexibility there).

app.rb
```diff
@@ -1,8 +1,19 @@
...
+    @all_peeps = Peep.all(params['filter_keyword'])
+    erb :messages
+  end
+
+  post '/messages/peep' do
+    new_peep = params[:new_peep]
+    Peep.create(new_peep) unless new_peep.empty?
+    redirect '/messages'
```
EdwardAndress, line 16: The restful routes for this app would be...

`get "/peeps"` and `post "/peeps"` to get all peeps and create a peep, respectively.

database_connection_setup.rb
```diff
@@ -0,0 +1,7 @@
+require './lib/database_connection'
```
EdwardAndress, line 1: Nice that you've abstracted the database connection.

spec/features/show_latest_first_spec.rb
```diff
@@ -0,0 +1,20 @@
+feature 'Show latest first' do
+  scenario 'A user can see the peeps in reverse chronological order' do
+    visit '/messages'
+    peep1 = 'Databases are complicated :('
+    peep2 = 'They are hard to understand!'
+    peep3 = 'There are so many of them!!!'
+    add_peep(peep1, Time.local(2021,11,5))
```
EdwardAndress, line 7: Nice use of a test helper here

views/messages.erb
```diff
@@ -0,0 +1,18 @@
+<h1>Chitter!</h1>
+<form action='messages/peep' method='post'>
+  <input type='text' name='new_peep' placeholder='New Peep'>
+  <input type='submit' value='Peep!'>
+</form>
+<form action='messages' method='get'>
```
EdwardAndress, line 6: Ah, I see that you added the option to show peeps in the two different orders.

spec/peep_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'peep'
+
```
EdwardAndress, line 2: To get 100% coverage, you'd also need tests for `.all` with filtering and `.toggle_order`

EdwardAndress:
Hey Patrick, this is a nice solution where you've clearly demonstrated the learning objectives.  Well done.  There are a few comments in line against your code. If you have any questions or would like to reply to my comments, please DM me on Slack :)

PR #139, opened 2021-11-05
Urs Burger Chitter Challenge

uburger:


app.rb
```diff
@@ -1,9 +1,23 @@
 require 'sinatra/base'
+require_relative './lib/message'
 
 class Chitter < Sinatra::Base
   get '/test' do
     'Test page'
   end
```
Jestfer, line 7: It's good to remove any extraneous code from your implementation before making a Pull Request - there'll be lots of testing and exploration as you work on something, but it tidies it up to remove anything that is no longer needed for the program to meet its specification.

Jestfer:
Nice work on this! You have partially demonstrated the learning objectives for this unit, by completing the first two user stories. Well done on your submission, there's a good demonstration of defining both GET and POST routes.

A good challenge to try out would be tackling the next user story. It would require some small changes to add the Time to your Model and to display it, but it makes for an interesting challenge when it comes to testing.

Well done!

PR #141, opened 2021-11-05
Added the example query into the sql file for table 01

MickStein:


spec/features/create_peep_page_spec.rb
```diff
@@ -0,0 +1,10 @@
+feature 'Posts new message' do
```
RumenKanev, line 1: I like that you have feature test. Testing is good 😀

spec/features/peeps_page_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  scenario 'user views added peeps' do
+    visit '/peeps'
+    # Add the test data
+    Peep.create(message: "Hello there!")
+    Peep.create(message: "General Kenobi")
+    Peep.create(message: "You are a bold one")
+
+    visit('/peeps')
```
RumenKanev, line 15: Not sure do you need line 15 as you already have it on line 9

app.rb
```diff
@@ -1,9 +1,26 @@
...
+  get '/peeps/new' do
+    erb :"peeps/new"
+  end
+
+  post '/peeps' do
+    Peep.create(message: params[:message])
+
+    redirect 'peeps'
+  end
+
```
RumenKanev, line 24: You have used 'post' to handle the data then redirect to get, which is the correct way of doing it. I'm guilty of handling all in the post request.

RumenKanev (commented):
Well done! 🙌 
I like you are following the correct process of handling the data from the view and committing regularly.

Katerina-codes:
- Good job Michael, 
- Nice use of restful routes in your app.rb
- This is a good attempt of this challenge. I can see you’ve completed the first few user stories and have also include tests for these. Nice work!
- Were there any missing tests that you wanted to include?
- What would your next steps be to continue with this challenge?

PR #142, opened 2021-11-05
DeanSmythe Chitter challenge. Initial PR

DeanSmythe:
Starting the PR with the first push.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,9 +1,16 @@
 require 'sinatra/base'
+require './lib/peeps'
 
 class Chitter < Sinatra::Base
   get '/test' do
     'Test page'
   end
+
+  get '/chitter' do
+    @peeps = Peeps.all
+    erb(:chitter)
+  end
+
 
   run! if app_file == $0
 end
--- a/lib/peeps.rb
+++ b/lib/peeps.rb
@@ -1,0 +1,9 @@
+require 'pg'
+
+class Peeps
+
+def self.all
+  return ['This is a peep!']
+end
+
+end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -98,4 +98,4 @@ RUBY VERSION
    ruby 3.0.2p107
 
 BUNDLED WITH
-   2.2.26
+   2.2.30
--- a/spec/features/viewing_chitter_spec.rb
+++ b/spec/features/viewing_chitter_spec.rb
@@ -1,0 +1,12 @@
+feature 'Viewing chitter page' do
+  
+  scenario 'visiting the chiter page' do
+    visit('/chitter')
+    expect(page).to have_content "Chitter"
+  end
+  scenario 'seeing a peep' do
+    visit('/chitter')
+    expect(page).to have_content "This is a peep!"
+  end
+
+end
--- a/views/chitter.erb
+++ b/views/chitter.erb
@@ -1,0 +1,8 @@
+<html>
+<h1> Chitter <h1>
+
+  <% @peeps.each do |peep| %>
+    <li> <%=peep%> </li>
+  <% end %>
+
+</html>


README.md
```diff
@@ -1,3 +1,23 @@
+##
```
rauldelahera, line 1: I like how you explain the functionality in your readme

lib/peeps.rb
```diff
@@ -0,0 +1,53 @@
...
+#   return results # Implicit return didnt; seem to work at some point
+# end
+
+  def self.add(peep_text:)
+    if ENV['ENVIRONMENT'] == 'test'
+      conn = PG.connect( dbname: 'chitter_test' )
+    else
+      conn = PG.connect( dbname: 'chitter')
+    end
+    p '+++++++'
```
rauldelahera, line 42: I'm confused about whether we are supposed to leave the print statements to get visibility in the code or not

views/add_peep_view.erb
```diff
@@ -0,0 +1,12 @@
+<html>
+
+<h1> Chitter </h1>
```
rauldelahera, line 3: simple and nice view :)

README.md
```diff
@@ -1,3 +1,23 @@
...
+Updates to the Readme.md by Dean Smythe
+See diagrams in Excel document and pdf.
+Chitter_Domain_Model.xlsx
+Chitter_Domain_Model.pdf
+
```
EdwardAndress, line 6: It's really great that you're using domain modelling as part of your planning process.  This will become particularly useful when you're working in teams (maybe you already discovered that this week!)

app.rb
```diff
@@ -1,9 +1,25 @@
 require 'sinatra/base'
+require 'sinatra/reloader'
+require './lib/peeps'
```
EdwardAndress, line 3: This all looks great!  The only thing I would suggest is that you revisit RESTful routing.  It's a really useful convention to follow and your future colleagues will probably expect you to do so.

```
```

README.md
EdwardAndress, line 6: It's great that you're using doman modelling and it'll become particularly powerful once you're working in a team.

lib/peep_support.rb
```diff
@@ -0,0 +1,10 @@
+
```
EdwardAndress, line 1: Good decision to make a method in a separate file for this!

lib/peeps.rb
```diff
@@ -0,0 +1,53 @@
+require 'pg'
+require_relative './peep_support'
+
+class Peeps
+
+  def self.all(order)
```
EdwardAndress, line 6: Nice that you can return peeps in either order.

lib/peeps.rb
EdwardAndress: Good question, @rauldelahera! Generally, these would be removed.

spec/features/adding_peep_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  scenario 'new peeps show a date' do
+    visit('/chitter')
+    click_button('Write peep!')
+    fill_in 'peep_text', with: "Another test peep"
+    click_button('Peep!')
+    expect(page).to have_content "#{Time.now.year}-#{Time.now.month}-#{check_leading_zero(Time.now.day)}"
+  end
+
+  scenario 'new peeps show in descending date order' do
```
EdwardAndress, line 22: Does this really test for descending order?

EdwardAndress (commented):
This is a nice solution!  Well done.  There are a few comments in line against the code. If you have any questions or want to reply, please DM me on Slack.

The next user story would be to add implement search (where you can find tweets by the presence of a particular word or words).  Do you know how you'd implement that?

PR #144, opened 2021-11-05
Initial set up

RJ21BL:
Added sinatra reloader, launchy and webrick gems.
Created chitter database and passed test run.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,1 +1,3 @@
 --require spec_helper
+--color
+--format documentation
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,9 @@ ruby '3.0.2'
 
 gem 'pg'
 gem 'sinatra'
+gem 'sinatra-reloader'
+gem 'launchy'
+gem 'webrick'
 
 group :test do
   gem 'capybara'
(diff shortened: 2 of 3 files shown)


app.rb
```diff
@@ -1,8 +1,17 @@
 require 'sinatra/base'
+require './lib/peep'
+require './database_connection_setup'
 
 class Chitter < Sinatra::Base
-  get '/test' do
-    'Test page'
+
```
oceansize, line 6: You're missing a route here. I notice you have a view file called `index.erb` that isn't being used.

To have your user arrive at that page, you need something like this:
```ruby
  get '/' do
    erb :index
  end
```

views/peeps.erb
```diff
@@ -0,0 +1,8 @@
+<% @peeps.each do |peep| %>
+  <li><%= peep.message %></li>
+<% end %>
+
+<form action="/create-peep" method='post'>
```
oceansize, line 5: You don't have a `post '/create-peep'` route in your controller (`app.rb`).

app.rb
```diff
@@ -1,8 +1,17 @@
...
+  get '/peeps' do
+    @peeps = Peep.all
+    erb :peeps
+  end
+
+  post '/send-peep' do
```
oceansize, line 12: If you rename `'/send-peep'` to `'/create-peep'` you should see what you're expecting.

PR #145, opened 2021-11-06
Chitter challenge - Jamie 

JRMBowers:


README.md
```diff
@@ -1,3 +1,18 @@
+## Personal Entry
+
```
EdwardAndress, line 2: Really nice reflection! :)

EdwardAndress: It's really cool that you're thinking about _when_ to refactor and reflecting on your decision there.  This will be really useful in the long term.

app.rb
```diff
@@ -1,9 +1,38 @@
 require 'sinatra/base'
+require_relative './model/message_handler'
```
EdwardAndress, line 2: This all looks great! The only thing I would suggest is that you revisit RESTful routing. It's a really useful convention to follow and your future colleagues will probably expect you to do so.

```ruby
```

app.rb
```diff
@@ -1,9 +1,38 @@
...
+    erb :filter
+  end 
+
+  post '/filter_reset' do 
+    redirect '/message_page'
+  end 
+
+  get '/message_page/filter' do 
+    @result = MessageHandler.all_filtered(filter: params[:filter])
+    erb :message_page
```
EdwardAndress, line 34: nice that you're re-using the message_page, rather than creating a specific page for filtered peeps.

model/database_handler.rb
```diff
@@ -0,0 +1,28 @@
+require 'pg'
+
+class Database
```
EdwardAndress, line 3: How would you adapt this class to handle more than one table / model?

spec/message_handler_spec.rb
```diff
@@ -0,0 +1,39 @@
+require_relative '../model/message_handler'
+require 'pg'
+
+describe MessageHandler do
+
+  describe '.all' do 
+    let(:db_entry) { ["id" => "1047", "messages" => "Hello World", "timestamp" => "05/11/2021"] }
```
EdwardAndress, line 7: Nice test isolation!

EdwardAndress:
Hi Jamie, overall this is a really nice solution and you've clearly demonstrated the learning objectives.  If you're looking to iterate on it, I would recommend that you revisit RESTful routing, as described in one of my comments.  The other thing which breaks convention is the naming of your templates.  These are minor things right now but will be important when you're working in a pro team.

PR #146, opened 2021-11-06
chitter_challenge

NenehB:


views/index.erb
```diff
@@ -0,0 +1,22 @@
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title> Chitter App</title>
```
RJ21BL, line 4: Nice clear index page with title.

app.rb
```diff
@@ -13,6 +13,13 @@ class Chitter < Sinatra::Base
     erb :index
   end 
 
- 
+  post '/messages' do 
+    p@message = params[:textarea]
```
RJ21BL: Nice use of p..is this for debugging?

database_connection_setup.rb
```diff
@@ -1,3 +1,3 @@
 require './lib/database_connection'
 
-DatabaseConnection.setup('chitter_test')
+DatabaseConnection.setup('chitter')
```
RJ21BL, line 3: Good use of separating responsabilities.

lib/peep_messages.rb
```diff
@@ -14,6 +14,10 @@ def self.all
       connection = PG.connect(dbname: 'chitter')
     end
 
+    p result = connection.exec("SELECT * FROM peeps")
+    result.map { | message | message['message'] }
+  #  p result = DatabaseConnection.query('SELECT * FROM peeps;')
```
RJ21BL: Maybe don't need the commented out code here?

app.rb
```diff
@@ -1,9 +1,24 @@
 require 'sinatra/base'
+require_relative './lib/peep_messages'
+require './database_connection_setup'
 
 class Chitter < Sinatra::Base
   get '/test' do
     'Test page'
   end
 
+  get '/' do
```
Katerina-codes, line 10: Regarding your routes names here, you could be more explicit. Instead of ‘/messages‘ - to be more consistent with the theme of the app you could have: 

**post /peeps do** 

Also, your **get '/' do** isn’t wrong but if you wanted to be more explicit you could do:

**get '/messages' do** or **get ‘/peeps’ do**

Hopefully that makes sense, if not then please do DM me to ask me questions :) 

More info on restful routes can be found here: https://github.com/makersacademy/course/blob/main/pills/rest.md 

Regarding TDD - I would have liked to have seen a test driven approach as it looks like you have tests missing, do you know have you could have approached this in a more test driven way? I’d like to know via DM please :D 

I can see that you’ve complete the first few user stories which is good. What would your next steps be to continue with this challenge? You can DM me with these if you’d like.

app.rb
```diff
@@ -1,9 +1,24 @@
...
     'Test page'
   end
 
+  get '/' do 
+     ENV
+    @peeps = PeepMessages.all
+    erb :index
+  end 
+
+  post '/messages' do
```
Katerina-codes, line 16: Regarding your routes names here, you could be more explicit. Instead of **‘/messages‘** - to be more consistent with the theme of the app you could have: 

**post /peeps do**

PR #148, opened 2021-11-07
chitter-challenge-rumen

RumenKanev:


MickStein:
Made great progress on the user stories. Good separation of MVC framework, clear variable/method naming and comment uploads to github. Good job at cleaning, assuming via Rubocop, some further cleaning up can be done but this is likely due to time and still progressing through the challenge. Great work!

spec/features/sort_peeps_spec.rb
```diff
@@ -0,0 +1,11 @@
+feature 'Sorting peeps by date' do
+  scenario "sort peeps" do    
+    visit("/home")
+    fill_in('message', with: "I'm a new peep 1")
+     fill_in('message', with: "I'm a new peep 2")
+    fill_in('message', with: "I'm a new peep 3")
+    click_button('Post Peeps')
+    click_button('New peeps on top')
+    expect(Message.all.first.message).to eq "I'm a new peep 3"
+  end
```
oceansize, line 10: Your test is giving you a false positive, because you only ever create the last message. It should be something like this:

```ruby
  scenario "sort peeps" do    
    visit("/home")
    fill_in('message', with: "I'm a new peep 1")
    click_button('Post Peeps')

    fill_in('message', with: "I'm a new peep 2")
    click_button('Post Peeps')

    fill_in('message', with: "I'm a new peep 3")
    click_button('Post Peeps')

    click_button('New peeps on top')
    expect(Message.all.first.message).to eq "I'm a new peep 3"
  end
```

lib/message.rb
```diff
@@ -0,0 +1,70 @@
...
+      else
+        Message.new(id: nil, message: 'Not a  Match', date: nil)
+      end
+    end
+           
+          
+         
+  end
+
+  def self.sort
```
oceansize, line 32: You have a bug in this method, which I only caught because I ran your app and could see in the browser the sort functionality wasn't working. Your tests were giving you a false positive, which leads me to believe you aren't in the habit of checking your work by actually running it in the browser.

If this is so, please be careful here. Tests are very valuable, but even they can lead you astray if they are set up incorrectly. A web developer should always check their software as if they were the user, in addition to running automated tests. This will help you not only catch more bugs, but help you find UX issues you might otherwise miss by just looking at your code.

PR #149, opened 2021-11-08
Chitter challeng in progress - 2 user stories completed

rauldelahera:
Currently working on User story #3. I'm trying to convert the type of message from a string to an object so we can easily add / access a second set of data in the object (date created)


DeanSmythe (commented):
Nice use of MVC.
Effective split of views to separate listing and adding a new message.

No diagrams? Could help to plan the next steps.
No changes to Readme.md. No additional instructions or description of the code function. Could help future reviewers.

app.rb
```diff
@@ -1,9 +1,29 @@
...
+    @peeps = Message.all
+    erb :'messages/index'
+  end
+  
+  get '/messages/new' do
+    erb :'messages/new'
+  end
+
+  post '/messages' do
+    Message.create(message: params[:message])
```
DeanSmythe, line 24: I also found that symbols (message:) need to be used to effectively pass the data into the method. But I don't know why.

DeanSmythe (commented):
Good use of both feature test and unit test to confirm the behaviour of the create metho.

DeanSmythe (commented):
Nice number of commits, showing progression and describing changes.

oceansize:
Hi Raul. I couldn't get your tests to run. I had to update the `Gemfile` to include `gem 'sinatra-contrib'` because you were getting the error:

```ruby
LoadError:
  cannot load such file -- sinatra/reloader
```

This fixed the issue for me.

app.rb
```diff
@@ -1,9 +1,29 @@
...
+require './lib/message'
 
 class Chitter < Sinatra::Base
+  configure :development do
+    register Sinatra::Reloader
+  end
+
   get '/test' do
     'Test page'
   end
```
oceansize, line 12: I understand why you would have a test page — but perhaps the next logical step would be to convert this to the homepage, with links to the 'get all messages', and 'create new message' routes.

PR #150, opened 2021-11-08
Apply database

jjegarcia:


app.rb
```diff
@@ -1,9 +1,30 @@
...
+  get '/' do
+    @list = PeepsData.list
+    erb :display
+  end
+  post '/message_box' do
+    PeepsData.add(Peep.new(params[:message]))
+    redirect '/'
+  end
+  post '/filter_box' do
+    $keyword = params[:keyword]
```
EdwardAndress, line 19: 👁️ – I see a global variable here!

app.rb
```diff
@@ -1,9 +1,30 @@
 require 'sinatra/base'
+require_relative './lib/peeps'
+require_relative './lib/peep'
+require_relative './lib/peeps_data'
```
EdwardAndress, line 4: This looks good.  Can you find a way of removing the global var?  Also, please be sure to revisit RESTful routing.

```
```

EdwardAndress:
Well done with your work on this – you've definitely demonstrated the learning objectives around reading from and writing to a database. I like the fact that you've also tried to separate concerns around Peeps but I'm not sure about this precise implementation.  If you want to separate the Peep business logic from data retrieval, you could use the repository pattern (which you have done, to some extent) but then you'd probably not need the `Peeps` class – that sorting behaviour could go in a `PeepRepository` class which also handled making DB queries.

Hope that make sense...  please DM me on Slack if you want to chat.

PR #151, opened 2021-11-08
Chitter Challenge - Mia

MiaGordon91:
I wasn't able to progress very far with the challenge, as I struggled to connect the test database with the feature tests.

I have created a partial solution for the second user story, but again, was not able to proceed any further due to the issue mentioned above.

I also think I started to refactor code too early as it led to further complications. 

I would like to identify where I may have gone wrong in the code and get clarity on how to use the test database in my code.


spdonovan (commented):
Looks like we both had the same issues! Test table not working either! Good tests though and easy to follow your code.

app.rb
```diff
@@ -1,9 +1,33 @@
...
+
+  get '/messages/new' do 
+    erb :'messages/new'
+  end 
+
+  post '/messages/new' do 
+    params 
   end
+    
+  # -- Unable to connect to test database
```
EdwardAndress, line 25: Did you get an error message when you tried to run the tests?

lib/message.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def self.all
+    if ENV['Environment'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+     else
+       connection = PG.connect(dbname: 'chitter')
+     end 
+
+    result = DatabaseConnection.query('SELECT * FROM peeps;')
```
EdwardAndress, line 18: It's a bit confusing that your Message class uses the peeps table.

EdwardAndress: Those of us who are used to the conventions get easily confused, perhaps :)

EdwardAndress:
Hey Mia, looks like you did some good work here but ran into some problems with testing?  Would you like to chat at some point and maybe we can figure out what was going on here?

PR #152, opened 2021-11-08
Rebecca Ransome- Chitter Challenge

Beckyrose200:


app.rb
```diff
@@ -1,9 +1,28 @@
...
+  post '/posting' do
+    Peeps.create(message: params[:posting])
+    redirect '/messages'
+  end 
+
   run! if app_file == $0
-end
+end 
+
+# As a Maker
```
oceansize, line 26: Try to get out of the habit of leaving comments in committed code :)

spec/chipping_spec.rb
```diff
@@ -0,0 +1,6 @@
+describe '.create' do
```
oceansize, line 1: What's `chipping`?

PR #153, opened 2021-11-08
Chitter Challenge

JackStorror:


Gemfile
```diff
@@ -10,6 +10,7 @@ gem 'sinatra'
 group :test do
```
JRMBowers, line 10: Noticed in commits: broken iterations uploaded, uploads not as frequent as they could be.

spec/peeps_spec.rb
```diff
@@ -0,0 +1,53 @@
+require 'peeps'
```
JRMBowers, line 1: could use doubling to remove coupling from tests

Gemfile
EdwardAndress: Very thorough reviewing, @JRMBowers!

EdwardAndress: Nice work

app.rb
```diff
@@ -1,9 +1,54 @@
 require 'sinatra/base'
+require './lib/peeps'
```
EdwardAndress, line 2: It would be worth revisiting RESTful routing. It's a really useful convention to follow and your future colleagues will probably expect you to do so.

```
```

app.rb
```diff
@@ -1,9 +1,54 @@
...
+require './lib/filter'
+require './database_connection_setup'
+require 'pg'
 
 class Chitter < Sinatra::Base
   get '/test' do
     'Test page'
   end
 
+  get '/chitter' do
```
EdwardAndress, line 14: There's a lot going on in this controller action.  Seems like you get all the peeps several times, which is confusing 🤔

views/chitter/index.erb
```diff
@@ -0,0 +1,30 @@
+<ul>
```
EdwardAndress, line 1: OK, looking at this I see why you've gathered a few different collections of peeps in your controller action.  you could simplify this template, by making it agnostic of the ordering, if you add some equivalent logic to your controller and / or Peep class.  For example, imagine you have a checkbox called `sort_by_newest_first`.  Then...

* In your controller you'd check the state of the checkbox
* If it's checked, you call something like Peep.all_reverse_chronological
* Else, you'd call Peep.all

lib/peeps.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def self.reverse
+    result = DatabaseConnection.query("SELECT * FROM peeps ORDER BY created_at DESC")
+    
+    result.map do |peep|
+      Peeps.new(id: peep['id'], message: peep['message'], created_at: peep['created_at'])
+    end
+  end
+
```
EdwardAndress, line 35: really awesome that you made it onto the stretch tasks

EdwardAndress:
Hey Jack!  Really well done here.  You've clearly exceeded the goals for the week, which is awesome.  I think the main focus here would be to reflect on code quality, particularly in your controller and views where there's some complexity to iron out.  I've left some comments in line against the code.  Please DM me on Slack if you have questions or wish to reply.

PR #154, opened 2021-11-08
Work in progress! Issues with test table

spdonovan:


MiaGordon91:
Great use of regular commits and commit messages

views/new.erb
```diff
@@ -0,0 +1,4 @@
+<form action="/messages" method="post">
+<input type="test" name="url" />
```
MiaGordon91: Would you consider renaming the name parameter with 'new_message' or 'new_post' instead of 'url'?

spec/features/web_helpers.rb
```diff
@@ -0,0 +1,5 @@
+def sign_in
```
MiaGordon91: Good use of a web_helper to avoid repetitive code

views/index.erb
```diff
@@ -0,0 +1,9 @@
+<h1 style="color: blue"> Welcome to Chitter!</h1>
```
MiaGordon91, line 2: Good use of headings to make the view more user friendly

app.rb
```diff
@@ -1,9 +1,25 @@
 require 'sinatra/base'
+require 'sinatra/reloader'
+require './lib/message'
 
 class Chitter < Sinatra::Base
   get '/test' do
     'Test page'
   end
```
Jestfer, line 8: It's good to remove any extraneous code from your implementation before making a Pull Request - there'll be lots of testing and exploration as you work on something, but it tidies it up to remove anything that is no longer needed for the program to meet its specification.

Jestfer:
Nice work on this! You have partially demonstrated the learning objectives for this unit, by completing the first two user stories. Well done on your submission, there's a good demonstration of defining both GET and POST routes.

A good challenge to try out would be tackling the next user story. It would require some small changes to add the Time to your Model and to display it, but it makes for an interesting challenge when it comes to testing.

Well done!

PR #155, opened 2021-11-08
chitter challenge 

Yasmin-jones:
Blocked majority of the project with passing feature 2 test


app.rb
```diff
@@ -5,5 +5,29 @@ class Chitter < Sinatra::Base
     'Test page'
   end
```
Jestfer, line 6: It's good to remove any extraneous code from your implementation before making a Pull Request - there'll be lots of testing and exploration as you work on something, but it tidies it up to remove anything that is no longer needed for the program to meet its specification.

Jestfer:
Nice work on this! You have partially demonstrated the learning objectives for this unit. Well done on your submission, there's a good demonstration of defining both GET and POST routes.

If you decide to come back to this, a good next step to try is to comment your functionality that is commented out back in and work on some unit tests as well for your app. Another good challenge to try out after that would be tackling the next user story. It would require some small changes to add the Time to your Model and to display it, but it makes for an interesting challenge when it comes to testing.

Well done!

PR #156, opened 2021-11-08
Chitter Challenge

antlerhead:


app.rb
```diff
@@ -1,9 +1,32 @@
 require 'sinatra/base'
+require 'sinatra/reloader'
+require './lib/message'
 
 class Chitter < Sinatra::Base
+  configure :development do
```
EdwardAndress, line 6: Nice RESTful routes! 🥳

spec/message_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'message'
+
+describe Message do
+  describe '.all' do
+    it 'returns all messages' do
+        connection = PG.connect(dbname: 'chitter_test')
```
EdwardAndress, line 6: nice that you're bypassing `.new` to test `.all` is isolation

views/messages/index.erb
```diff
@@ -0,0 +1,5 @@
+<ul>
+  <% @messages.each do |message| %>
+    <li><%= message %></li>
+  <% end %>
+</ul>
```
EdwardAndress, line 5: It's not in the requirements but it might be nice to include a link to the new message form ;)

EdwardAndress:
Hey Max, this is a nice solution which clearly demonstrates the learning objectives around reading from and writing to database.  So, well done there!

I've left a couple of comments inline against your code.  Also, I had a look your commits (or at least the commit messages) and it looks like you took a nice methodical approach here.  I particularly like the fact that you started by returning hard coded data.  This approach of building something by taking small steps will serve you well.

I believe the next user story would be to add a date stamp so that you can see when a message was posted and the one after that would be to sort them in reverse chronological order.

Do you have an idea about how to do these things?

PR #157, opened 2022-01-21
Marcus' Janky Chitter

marcusventin:


lib/peep.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def self.search(keyword)
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
+
+    result = connection.exec_params("SELECT * FROM peeps WHERE (message)
```
JALWilliams, line 49: I think you could have just used the "exec" because you wouldn't be able to have a SQL injection here with the query as there's no insert.

spec/features/search_peeps_spec.rb
```diff
@@ -0,0 +1,20 @@
+
+feature 'search by keyword' do
+  scenario 'it filters by keyword' do
+    add_peep('breaking news', 'I made eggs today')
```
JALWilliams, line 4: really clean how you've used your methods instead of SQL directly in your feature tests.

spec/features/web_helpers.rb
```diff
@@ -0,0 +1,6 @@
+def add_peep(username, message)
```
JALWilliams, line 1: Very clean & refactored! I should've done this too.

views/index.erb
```diff
@@ -0,0 +1,43 @@
...
+
+  <div id="new search">
+    <form action='/peeps/search' method='post'>
+      <input type='text' placeholder='Search by keyword' name='search_keyword'>
+      <input type='submit' value='Search'>
+    </form>
+  </div>
+
+  <div id="list of peeps">
+    <% if @peeps.empty? %>
```
JALWilliams, line 28: This is great! shows that you're thinking about the user experience

views/results.erb
```diff
@@ -0,0 +1,44 @@
...
+  <div id="search heading">
+    <div id="search feedback">
+      <% if @search.length == 1 %>
+      <h3 style="font-family: helvetica">Your search for <%= @keyword %> returned 1 result.</h3>
+      <% else %>
+      <h3 style="font-family: helvetica">Your search for <%= @keyword %> returned <%= @search.length %> results.</h3>
+      <% end %>
+    </div>
+    <div id="new search">
+      <h3 style="font-family: helvetica">Search again?</h3>
```
JALWilliams, line 20: Great to have a loop in your query for the next step post search results!

spec/peep_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      expect(test_peep).to be_a Peep
+      expect(test_peep.username).to eq 'test user'
+      expect(test_peep.message).to eq 'test message'
+    end
+  end
+
+  describe '.search' do
+    it { is_expected.to respond_to(:search).with(1).argument }
+
+    it 'returns peeps with the designated keyword' do
```
JALWilliams, line 20: Really thorough test coverage! :)

lib/peep.rb
```diff
@@ -0,0 +1,56 @@
+require 'pg'
+
+class Peep
+  attr_reader :id, :username, :message, :datetime
+
+  def initialize(id:, username:, message:, datetime:)
+    @id = id
+    @username = username
```
JALWilliams, line 8: Think that this is a great addition to the bot - would be more reflective of Twitter

views/results.erb
```diff
@@ -0,0 +1,44 @@
...
+</style>
+
+<div id="body">
+  <div id="heading">
+    <h1 style="font-family: helvetica">CHITTER</h1>
+  </div>
+  <div id="search heading">
+    <div id="search feedback">
+      <% if @search.length == 1 %>
+      <h3 style="font-family: helvetica">Your search for <%= @keyword %> returned 1 result.</h3>
```
JALWilliams, line 14: I like how you've provided context on the search page for the term they searched for than just the results directly.

README.md
```diff
@@ -1,5 +1,90 @@
...
+> As a Maker,  
+> So that I can find relevant peeps,  
+> I want to filter on a specific keyword.
+
+## Features
+Users can submit peeps to the message board, where they are attributed to a user's name. Once submitted, peeps are presented in the order of most recent first.  
+Users can also search the peeps database by keyword. Thanks to some snazzy coding, the search function is case-insensitive.
+
+## How to Use
+
```
EdwardAndress, line 34: Nice that you've included this helpful guidance and screenshots

app.rb
```diff
@@ -1,9 +1,33 @@
 require 'sinatra/base'
+require './lib/peep'
 
 class Chitter < Sinatra::Base
+  enable :sessions
+
   get '/test' do
```
EdwardAndress, line 7: You can probably remove the `/test` route

app.rb
```diff
@@ -1,9 +1,33 @@
...
+  get '/peeps' do
+    @peeps = Peep.all
+    erb(:index)
+  end
+
+  post '/peeps' do
+    Peep.add(username: params[:peep_user], message: params[:peep_message])
+    redirect '/peeps'
+  end
+
```
EdwardAndress, line 20: There is a way of adding search functionality to the index action `GET "/posts"`.  It would look something like this...

```ruby
get "/posts" do
  if params[:search_keyword]
    @peeps = Peep.search(params[:search_keyword])
  else
    @peeps = Peep.all
  end

  erb :index
end
```

You could argue that this breaks SRP and you might have a point. I think, however, I'm OK with it since either way the user ends up with a list of some `Peep`s.

Alternatively, you could still simplify things a little bit by using the same `index` template for both search results and showing all peeps.

Either way, I think the request generated when someone submits the search form would be a GET request, since nothing is being added to the DB. It's just a GET request with some params. Similar to if you wanted to see one single peep, which would be found using it's ID, passed in the URL.

db/migrations/01_create_chitter_table.sql
```diff
@@ -1 +1 @@
-CREATE TABLE peeps(id SERIAL PRIMARY KEY, message VARCHAR(60));
+CREATE TABLE peeps (id SERIAL PRIMARY KEY, username VARCHAR(60), message VARCHAR(200));
```
EdwardAndress, line 1: I can't add comments to the diagrams themselves, but they're great!

lib/peep.rb
EdwardAndress: Give it a try :)

I think that the key is whether or not user input is incorporated into the query, since the intended query will be cut short by the SQL that a nefarious user injects.

So I think it's justified but... try it out! :)

spec/features/search_peeps_spec.rb
EdwardAndress: Agreed!

EdwardAndress:
Hey Marcus, this is a great solution which clearly demonstrates the learning objectives. I've left a couple of comments in line against the code. Beyond those, the thing to reflect on is that there's some repetition when it comes to connecting to the database. In Bookmark manager we used a databaseconnection object to encapsulate that logic – though I appreciate that you may not have had time to do that here!

PR #159, opened 2022-01-21
Orla Dunlop Chitter Challenge

odunlop:


lib/db_connection.rb
```diff
@@ -0,0 +1,11 @@
...
+
+class DBConnection
+  def self.setup(dbname)
+    @connection = PG.connect(dbname: dbname)
+  end
+
+  def self.query(sql, params = [])
+    @connection.exec_params(sql, params)
+  end
+end
```
siellsiell, line 11: A nice way to reduce code duplication :)

db/migrations/01_create_chitter_table.sql
```diff
@@ -1 +1,2 @@
 CREATE TABLE peeps(id SERIAL PRIMARY KEY, message VARCHAR(60));
+ALTER TABLE peeps ADD COLUMN date timestamp with time zone;
```
siellsiell, line 2: Nice use of the `DATE` datatype.

lib/peep.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def self.create(message = "default peep", date = Time.now)
+    new_peep = DBConnection.query(
+      "INSERT INTO peeps (message, date) VALUES($1, $2) RETURNING  id, message;", [message, date]
+    )
+    Peep.new(id: new_peep[0]['id'], message: new_peep[0]['message'], date: new_peep[0]['date'])
+  end
+
+  def self.chronological
```
siellsiell, line 29: Minor: this says chronological but the result is actually the reverse of chronological

spec/features/viewing_peeps_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  scenario 'The /peeps page displays all peeps' do
+    add_row_to_test_database
+
+    visit '/peeps'
+    
+    expect(page).to have_content "This is a peep!"
+  end
+
+  scenario 'The user can see the date the peep was posted' do 
+    add_row_to_test_database
```
siellsiell, line 13: Minor: Using a helper method here helps to reduce duplication. One drawback of helper methods though is that they can make the test harder to follow as a reader of the test doesn't know at first glance what kind of data is being added to the database.

A compromise could be to make your helper method take an argument, e.g. the text of the peep so that you can call it like this `add_row_to_test_database("some text")`. That would make the link between line 17 and line 13 clearer.

app.rb
```diff
@@ -1,9 +1,31 @@
...
+    erb(:index)
+  end
+
+  get '/peeps/new' do 
+    erb(:new)
+  end
+
+  post '/peeps' do
+    Peep.create(params['message'])
+    redirect '/peeps'
```
siellsiell, line 27: Good use of redirect

siellsiell (commented):
Hey Orla,

Thanks for submitting your code! You've got a clean well-tested solution there and your diagram shows a solid understanding of the MVC architecture. I've left you a few comments for things you can do improve it further, most of them minor.

Let me know if you have any questions :)

PR #160, opened 2022-01-21
Johnny Young Chitter Challenge

jyoung-gds:
User Stories:
- As a Maker, so that I can see what people are doing, I want to see all the messages (peeps) in a browser

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,8 +1,20 @@
 require 'sinatra/base'
+require 'sinatra/reloader'
+require_relative './lib/peep'
 
 class Chitter < Sinatra::Base
+
+  configure :development do
+    register Sinatra::Reloader
+  end
+
   get '/test' do
     'Test page'
+  end
+
+  get '/peeps' do
+    @peeps = Peep.all
+    erb :index
   end
 
   run! if app_file == $0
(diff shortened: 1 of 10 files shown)


adamhind:
### Positives 

# Keyword.downcase
Thoughtful use for filtering. 
# Ordering for Peeps
Good ordering using the date in descending order from the SQL query. 
# Refactoring 
All of your code is refactored with SPEC helper being used to run two methods that truncate and add rows to the table. 
# README 
Your domain table and migrations are extremely helpful for new devs. Your code overall was easy to follow and named very well. 

### Areas of development
# Security 
In your filter query you are open to SQL injection. 
# User experience 
An add button might be nice on the index.erb to take you to add a new peep. 

Very nice Johnny!

app.rb
```diff
@@ -1,9 +1,38 @@
...
+require 'sinatra/reloader'
+require_relative './lib/peep'
 
 class Chitter < Sinatra::Base
+
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  get '/' do
```
Katerina-codes, line 11: It's a bit strange to have a route redirecting to another route, since we don't ask for a homepage, you can change this so we're just getting all peeps.

PR #161, opened 2022-01-21
Starting the challenge

JennyEMorrison:
Made and added a basic domain model for this challenge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,5 +63,9 @@ I want to see a list of peeps in reverse chronological order
 As a Maker
 So that I can find relevant peeps
 I want to filter on a specific keyword
-```<img width="758" alt="Screenshot 2022-01-21 at 10 14 14" src="https://user-images.githubusercontent.com/94904399/150512927-445bb64e-9d61-4968-a83f-855c4b860607.png">
+```
+
+# Below is a basic domain model for this challenge
+
+<img width="758" alt="Screenshot 2022-01-21 at 10 14 14" src="https://user-images.githubusercontent.com/94904399/150512927-445bb64e-9d61-4968-a83f-855c4b860607.png">
 
--- a/README.md
+++ b/README.md
@@ -63,4 +63,5 @@ I want to see a list of peeps in reverse chronological order
 As a Maker
 So that I can find relevant peeps
 I want to filter on a specific keyword
-```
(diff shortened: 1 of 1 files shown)


lib/peeps.rb
```diff
@@ -1,9 +1,30 @@
...
       "This is my first peep",
       "This is my second peep",
-      "This is my third peep"
+      "This is my third peep",
+      "peep"
     ]
   end
+
+  def self.create(message:)
+    if ENV['ENVIRONMENT'] == 'test'
```
MichaelBodendorf, line 22: You have mapped the environments in your code very well

spec/peeps_spec.rb
```diff
@@ -1,13 +1,24 @@
...
-    it 'returns all peeps' do
-      peeps = Peep.all
-
-      expect(peeps).to include("This is my first peep")
-      expect(peeps).to include("This is my second peep")
-      expect(peeps).to include("This is my third peep")
-    end
+  it 'returns all peeps' do
+    connection = PG.connect(dbname: 'chitter_test')
+
```
MichaelBodendorf, line 6: some DRY here ;-)

app.rb
```diff
@@ -1,9 +1,28 @@
 require 'sinatra/base'
+require './lib/peeps'
 
 class Chitter < Sinatra::Base
   get '/test' do
```
Katerina-codes, line 5: I don't think you need this test page any more so it can be deleted :)

app.rb
```diff
@@ -1,9 +1,28 @@
...
     'Test page'
   end
 
+  get '/' do
```
Katerina-codes, line 9: Since the first user story is that you can see all peeps, there's no need to create a homepage here.

lib/peeps.rb
```diff
@@ -0,0 +1,30 @@
...
+  def self.all
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end  
+    result = connection.exec("SELECT * FROM peeps;")
+    result.map { |peep| peep['message'] }
+
+    @peeps = [
```
Katerina-codes, line 13: You get the peeps from the database on lines 10 - 11 but then you don't use them because @peeps (on line 13) is hardcoded. You should use result instead and assign that to **@peeps** :)

JennyEMorrison:
Thanks for the feedback Katerina! I got a little bit lost by the end (as I usually do 😅)

Always good to know where I can alter things!

Thanks again,
Jenny

________________________________

PR #162, opened 2022-01-21
Sam Pickering

samhpickering:


app.rb
```diff
@@ -3,6 +3,8 @@
 require_relative 'database_setup'
 
 class Chitter < Sinatra::Base
+  MAX_PEEP_LENGTH = 281
```
jackiebenn, line 6: I like that you have made use of a constant to define the max peep length

jackiebenn:
Clear and easily readable code

PR #163, opened 2022-01-21
Alex Wood PR

alexwood91:


lib/peep.rb
```diff
@@ -0,0 +1,13 @@
+require 'pg'
+
+class Peep
+  def self.all 
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
```
Aadam-Ali, line 9: Extract database connection to it's own class.

siellsiell (commented):
Hey it's Simo, a coach :)

Thanks for submitting your challenge code and apologies for the slow response. This was a good start! I like that you drew a diagram of your architecture, it's a good way to plan and also a good way to illustrate your architecture for reviewers of the code.

I saw that you got started with a `creating_peeps_spec.rb` test -- did you run out of time or get blocked? It would be good to revisit this with a coach if there is anything that stumped you.

I have a few minor comments below.

PR #164, opened 2022-01-21
Michael Pull Request

michael-zwickler:
All functionality added. Filter option currently having the risk of SQL injection...
>>> Filter function updated with exec_params and $1


app.rb
```diff
@@ -1,9 +1,25 @@
 require 'sinatra/base'
+require './lib/message'
 
 class Chitter < Sinatra::Base
-  get '/test' do
-    'Test page'
+
```
EdwardAndress, line 5: There is a way of adding search functionality to the index action GET "/posts". It would look something like this...

```ruby
get "/posts" do
  if params[:filter]
    @peeps = Peep.search(params[:filter])
  else
    @peeps = Peep.all
  end

  erb :index
end
```
You could argue that this breaks SRP and you might have a point. I think, however, I'm OK with it since either way the user ends up with a list of some Peeps.

Alternatively, you could still simplify things a little bit by using the same index template for both search results and showing all peeps.

Either way, I think the request generated when someone submits the search form would be a GET request, since nothing is being added to the DB. It's just a GET request with some params. Similar to if you wanted to see one single peep, which would be found using it's ID, passed in the URL.

app.rb
```diff
@@ -1,9 +1,25 @@
...
+  enable :sessions
+
+  get '/' do
+    @filter_tag = session[:filter]
+    @peeps = Message.all(filter: @filter_tag)
+    erb(:board)
   end
 
+  post '/peep/new' do
```
EdwardAndress, line 14: The RESTful route for creating a peep would be `POST "/peeps"`

lib/message.rb
```diff
@@ -0,0 +1,36 @@
...
+class Message
+
+  attr_reader :id, :message, :time
+
+  def initialize(id:, message:, time:)
+    @id = id
+    @message = message
+    @time = time
+  end
+
```
EdwardAndress, line 12: I think I would prefer to put the raw time into the db.  It's a bit more flexible and if you want to search or filter on this field later you don't need to do any parsing.

Converting the time to a user friendly format is presentational and could be done in the view, or by a specific presenter class.

Doing this also means that if you change your mind about the design of the user interface (and the way that times are displayed there) you don't need to change everything in the posts table :)

lib/message.rb
```diff
@@ -0,0 +1,36 @@
...
+    time = Time.now.strftime("Posted on %d.%m.%Y at %I:%M %p")
+    connect_to_db.exec_params("INSERT INTO peeps(message,time) VALUES($1, $2);", [message,time])
+  end
+
+  def self.all(filter: '')
+    return filter_by(tag: filter) unless (filter.nil? || filter.empty?)
+    result = connect_to_db.exec("SELECT * FROM peeps ORDER BY id DESC;")
+    result.map { |peep| Message.new(id: peep['id'], message: peep['message'], time: peep['time']) }
+  end
+
```
EdwardAndress, line 23: Nice that you're thinking about using private methods

EdwardAndress:
Hi Michael, this is a really nice solution which clearly demonstrates the learning objectives. I've left a few comments in line against the code. I hope they all make sense and are useful :)

PR #165, opened 2022-01-21
Adam Hindmarsh's Chitter Chitter Bang Bang 

adamhind:


jyoung-gds:
Good stuff:
- code works and all tests pass!
- code is well segmented with good use of single responsibility principle
- code is excellently formatted with no rubocop offenses
- good use of domain model, make sure to include it in your README next time
- excellent use of CSS to make it look good and good user interface

Improvements:
- wrap connection into separate class for DRY
- could use separate methods for SQL queries? Not sure if that’s necessary though
- potentially move all database set up (i.e. creating test posts) to your setup_test_database.rb add_row_to_test_database method, and run that in the spec_helper.rb so it runs before every test. Keeps all test posts the same across all tests so you know what you’re testing for.
- RESTful routes could be used for search and create peeps
- self.search susceptible to SQL injection
- add DB migration for adding time column to DB and include this in README.md

lib/peep.rb
```diff
@@ -0,0 +1,53 @@
...
+      Peep.new(id: peep['id'], message: peep['message'], time: peep['time'])
+    end 
+  end
+
+  def self.post(message:)
+    if ENV['ENVIRONMENT'] == 'test'
+      con = PG.connect(dbname: 'chitter_test')
+    else
+      con = PG.connect(dbname: 'chitter')
+    end
```
siellsiell, line 33: This piece of code is repeated several times in this file. How could you  reduce this duplication (to make this code easier to maintain)?

app.rb
```diff
@@ -1,9 +1,28 @@
 require 'sinatra/base'
+require_relative './lib/peep'
 
 class Chitter < Sinatra::Base
-  get '/test' do
-    'Test page'
+
+  enable :sessions, :method_override
+
+  get '/messageboard' do
```
siellsiell, line 8: It might be worth having an index route as well as that's where users will land by default on a website. You could connect it to  `/messageboard` using a redirect if that's where you want users to end up.

views/messageboard/results.erb
```diff
@@ -0,0 +1,25 @@
...
+    <h1> Here are the results </h1>
+    <br>
+    <br> 
+    <ul>
+      <% @peeps.each do |peep| %>
+        <li>
+            <%=  peep.message  %> 
+            <%=  peep.time  %>
+        </li>
+      <% end %>
```
siellsiell, line 22: Is there a way you could combine this view with the `views/messageboard/index.erb` view so that you can reuse some of the shared code for displaying peeps? 

Yours is a fair solution and implementing what I suggest above might require some refactoring of your model and controller code to, so I'm only including this here as food for thought.

PR #167, opened 2022-01-21
Theresa Okyere update so far

TheresaOkyere:


app.rb
```diff
@@ -1,9 +1,36 @@
 require 'sinatra/base'
+require 'sinatra/reloader'
+require './lib/message'
 
 class Chitter < Sinatra::Base
+  configure :development do
+    register Sinatra::Reloader
+  end
   get '/test' do
```
EdwardAndress, line 9: You can probably get rid of this `/test` stuff :)

spec/features/creating_peeps_spec.rb
```diff
@@ -0,0 +1,10 @@
+feature 'Adding a new message(peep)' do
+  scenario 'A user can post a message(peep) to chitter' do
+    visit('/messages/new')
+    fill_in('message', with: 'a new post')
+    fill_in('date-posted', with: '2022-01-12T19:30')
```
EdwardAndress, line 5: Looks like this date-posted doesn't get stored in the database – is that something you were in the process of implementing?

lib/message.rb
```diff
@@ -0,0 +1,24 @@
+require 'pg'
+
```
EdwardAndress, line 2: Everything looks great here! The next iteration could involve trying to reduce some of the duplication.

spec/message_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+      connection.exec("INSERT INTO peeps (message) VALUES('How are you today?');")
+
+      messages = Message.all
+
+      expect(messages).to include('Hi there')
+      expect(messages).to include('How are you today?')
+    end
+  end
+  
+describe '.create' do
```
EdwardAndress, line 18: What happened to the indentation? :)

views/homepage.erb
```diff
@@ -0,0 +1,11 @@
+
+<div class="title">
+
+<h1 style="text-align:center"> Welcome to peep!</h1>
+</div>
+
+<form style="text-align:center">
+    <button type="submit" class="btn btn-primary btn-lg btn-block" formaction="http://localhost:9292/messages">view peeps!</button>
```
EdwardAndress, line 8: here you've included `localhost` in your formaction, which means it'll only work when the client and server are on the same machine (as is the case when you're developing the app). If you were to deploy it to the cloud, this would break.

EdwardAndress:
Hi Theresa, this is a nice solution and you've clearly demonstrated the learning objectives – reading from and writing to a database. The main thing to reflect on is the code repetition.

PR #168, opened 2022-01-21
Done up to 4th user story

alicegray33:


lib/peep.rb
```diff
@@ -0,0 +1,41 @@
...
+      Peep.new(id: peep['id'], message: peep['message'], date: peep['date'])
+    end   
+  end
+
+  def self.create(message:)
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
```
siellsiell, line 35: This piece of code appears twice in this file. How could you reduce this duplication?

db/migrations/01_create_chitter_table.sql
```diff
@@ -1 +1,3 @@
 CREATE TABLE peeps(id SERIAL PRIMARY KEY, message VARCHAR(60));
+
+ALTER TABLE peeps ADD date BIGINT;
```
siellsiell, line 3: Postgres has a more natural datatype for representing dates and times (https://www.postgresql.org/docs/9.1/datatype-datetime.html) and Ruby has a [DateTime class](https://ruby-doc.org/stdlib-2.6.1/libdoc/date/rdoc/DateTime.html) which is a bit easier to work with for this purpose than Timestamp.

This would also allow you to simplify your date-conversion code in the Peep class :)

spec/peep_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+describe Peep do
+
+  describe '.all' do
+    it 'returns all peeps' do
+      connection = PG.connect(dbname: 'chitter_test')
+      connection.exec("INSERT INTO peeps (message) values ('This is a peep!');")
+
+      peeps = Peep.all
+
+      expect(peeps[0].message).to include 'This is a peep!'
```
siellsiell, line 12: I think you can make a stronger assertion than that. The value of `message` should be equal to `"This is a peep!"`, not just include it.

spec/features/create_peep_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+  scenario 'add a peep and see the date it was posted' do
+    # So Date/times should be: 21/01/2022 at 14:06
+    allow(Time).to receive(:now).and_return(1_642_774_001)
+    visit('/new')
+    fill_in('message', with: 'This is also a Peep')
+    click_button('Peep it!')
+    expect(page).to have_content 'This is also a Peep'
+    expect(page).to have_content 'Posted: 21/01/2022 at 14:06'
+  end
+end
```
siellsiell, line 17: Nice feature test!

app.rb
```diff
@@ -1,9 +1,25 @@
...
+    erb :index
+  end
+
+  get '/new' do
+    erb :new
+  end
+
+  post '/new' do
+    Peep.create(message: params[:message])
+    redirect '/'
```
siellsiell, line 21: Good use of redirect :)

README.md
```diff
@@ -1,36 +1,19 @@
...
 
-## Set up
+This task started relatively easily and similar to the bookmarks pairing challenge we have been working on. After setting up the development and test database and installing the required gems I started with a feature test aimed at the first user story of seeing all messages. From here it developed to unit testing the Peep object that could retreive all peeps from the database and then creating an erb file to display them. It was a similar case for the second user story which led me to do more unit testing on the Peeps class so it could create a new peep.
 
-To setup the database:
+The user story to see the date made things a bit more complicated but as before I started with the feature test which led to me updating the database. I initially added a date column with the timestamp data type so it would automatically input the date when a new record was created. However I struggled to think of a way to test that accurately and so I decided to change the datatype to a bigint and save the time as "number of seconds since the unix epoch." I felt that gave me more flexibility when it came to formatting the date, and meant I could test it using rspecs `expect ` call.
```
siellsiell, line 9: It can be tricky to write tests that rely on timing. You took the right approach by using mocking to control what time would be returned in a test. Lower down in another comment I gave some suggestions for other data types you could have used other than timestamp. 

Using a DATE or DATETIME column in postgres is preferable because you can then rely on the database to catch bad data being inserted. Using BIGINT means the DB can't detect if your code accidentally inserted numbers that are not valid timestamps for example.

README.md
```diff
@@ -1,36 +1,19 @@
...
-* Run the query we have saved in the file 01_create_chitter_table.sql
-
-* `bundle install`
-* `rspec`
-
-You should see 1 passing test.
+![All Peeps](https://github.com/alicegray33/chitter-challenge-apprenticeships/blob/main/docs/allpeeps.png?raw=true)
+![New Peep](https://github.com/alicegray33/chitter-challenge-apprenticeships/blob/main/docs/newpeep.png?raw=true)
 
 ## User stories
```
siellsiell, line 19: Nice readme! Your app looks great too :)

README.md
```diff
@@ -1,36 +1,19 @@
...
 
-* Connect to psql
-* Create the database using the psql command `CREATE DATABASE chitter;`
-* Connect to the database using the psql command `\c chitter`;
-* Run the query we have saved in the file 01_create_chitter_table.sql
-* Populate your table with a row by running `INSERT INTO peeps (message) values ('This is a peep!');`
+For the fourth user story of displaying the peeps in reverse chronological order: I found myself in a position where I didn't write a test first. It just meant changing the SQL query to the database to `ORDER BY date DESC` with nothing else needed changing. I ran out of time before figuring out how to write a test to see if the peeps were actually appearing in descending order but the current tests pass with 97.22% coverage. And the only reason they aren't 100% is due to the `if` statements that choose which database to use.
```
siellsiell, line 11: Do you think you'd have an idea of how you'd approach writing the tests for the fourth user story?

siellsiell (commented):
Hey it's Simo, a coach :)

Thanks for submitting your challenge code and apologies for the slow response. This was good solution to the challenge! I left some comments with more details :) Do let me know if you have any questions.

spec/peep_spec.rb
siellsiell: You might also want to test the `Peep#date` method

PR #169, opened 2022-01-22
complete

Tom-V1:


app.rb
```diff
@@ -1,9 +1,54 @@
...
+
+  get '/new' do
+    erb :new
+  end 
+
+  get '/search' do
+    erb :search
+  end 
+
+  get '/search_result' do
```
EdwardAndress, line 27: Do you _have_ to have a route specifically for this?  Could you combine it with `/`?

app.rb
```diff
@@ -1,9 +1,54 @@
...
+    @keyword = session[:keyword]
+    @messages = Message.search(search: @keyword)
+    p @keyword
```
EdwardAndress, line 30: Watch out for these lingering debug lines :)

lib/db.rb
```diff
@@ -0,0 +1,12 @@
+require 'pg'
+
```
EdwardAndress, line 2: nice that you did this as it makes the codebase a lot DRYer

lib/message.rb
```diff
@@ -0,0 +1,45 @@
...
+    result = connection.exec("SELECT * FROM peeps")
+    result.map do |message|
+      Message.new(id: message['id'], message: message['message'], date: message['date'])
+    end
+     
+  end
+
+  def self.create(message:) 
+    connection = Db.connect
+    connection.exec_params("INSERT INTO peeps (message,date) VALUES($1,current_timestamp);",[message])
```
EdwardAndress, line 25: I wonder if you can find a neat way of pushing all the SQL into your Db class....

spec/features/search_message_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+    Message.create(message: 'blahh')
+    Message.create(message: 'test_2')
+    Message.create(message: 'rahhh')
+
+    visit('/')
+    click_button('Search')
+    fill_in('keyword', with: "test")
+    click_button('Search')
+    
+    expect(page).to have_content("test_2")
```
EdwardAndress, line 12: Maybe you could add another expectation here which asserts that the other posts are filtered out

spec/features/sort_message_spec.rb
```diff
@@ -0,0 +1,13 @@
+feature 'Sort messages' do
+  scenario 'Shows all bookmarks sorted by date' do
+    Message.create(message: 'test_1')
+    Message.create(message: 'test_2')
+    Message.create(message: 'test_3')
+
+    visit('/')
+    click_button('Sort by date')
+    
+    expect(page).to have_content("test_1")
```
EdwardAndress, line 10: This feature tests is a little tricky but, using capybara, you can grab all elements by, for example, class and that will return an array of elements in the same order as they appear on the page.  So then you can test for order.

spec/unit/message_spec.rb
```diff
@@ -0,0 +1,52 @@
+require 'message'
+
+describe Message do
+  describe '.all' do
+    it 'returns message' do
+      Message.create(message: 'This is the a message')
+
+      message = Message.all
```
EdwardAndress, line 8: I'd probably pluralise the variable name here since it Message.all returns an array

spec/unit/message_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+
+      expect(message.first).to be_a Message
```
EdwardAndress, line 10: Do you really care if the object is of type Message?  What the program were expanded to include some other type of post?  Maybe you only really care that they all behave in the right way...

spec/unit/message_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  describe '.search' do
+    it 'Sorts messages by date' do
+      Message.create(message: 'first message')
+      Message.create(message: 'second message')
+      Message.create(message: 'third message')
+
+      message = Message.search(search: 'second')
+
+      expect(message.first.message).to eq ('second message')
```
EdwardAndress, line 48: Can you go further and assert that some messages are filtered out?

lib/message.rb
```diff
@@ -0,0 +1,45 @@
...
+    @id = id
+    @message = message
+    @date = date
+  end
+
+  def self.all
+    connection = Db.connect
+    result = connection.exec("SELECT * FROM peeps")
+    result.map do |message|
+      Message.new(id: message['id'], message: message['message'], date: message['date'])
```
EdwardAndress, line 18: This business of wrapping return values in instances of Message requires a bit of mental gymnastics. Well done for doing that.

views/index.erb
```diff
@@ -0,0 +1,22 @@
+
+<% @messages.each do |message| %>
+  <ul style="list-style-type:none;">
+    <li> 
+      <h2><%= message.message%></h2>
+      Posted at <%= message.date[0...-10]%>
```
EdwardAndress, line 6: perhaps you could add a method to Message called something like `pretty_date` which takes care of the formatting.  Or, even better, you could create a DatePresenter class, with a `pretty_date` that takes a raw date and then returns a pretty one.

views/search_result.erb
```diff
@@ -0,0 +1,17 @@
+<% @messages.each do |message| %>
```
EdwardAndress, line 1: This template is very similar to index.erb – can you combine them into one?

views/sorted.erb
```diff
@@ -0,0 +1,13 @@
+<% @messages.each do |message| %>
```
EdwardAndress, line 1: This template is similar to sear_result and index

EdwardAndress:
Hey Tom! Really great work here – you've clearly demonstrated the learning objectives.  There's a bunch of little things that could be done to improve code quality and I've mentioned those inline against the code.  The most important one is to ensure things are really well tested.

PR #173, opened 2022-01-24
Jackie - Chitter

jackiebenn:


views/add.erb
```diff
@@ -0,0 +1,6 @@
+<form action='/peeps/add' method='post'>
```
samhpickering, line 1: To follow RESTful routing it might be better to POST directly to /peeps

app.rb
```diff
@@ -1,9 +1,24 @@
 require 'sinatra/base'
+require './lib/peep'
 
 class Chitter < Sinatra::Base
   get '/test' do
```
Katerina-codes, line 5: I don't think you need this so it can be deleted :)

PR #174, opened 2022-01-24
Michael B. chitter

MichaelBodendorf:


JennyEMorrison:
I find the code very readable and well presented. You managed to get a lot done as well and really interesting to see your approach.

app.rb
```diff
@@ -1,9 +1,29 @@
...
   get '/test' do
     'Test page'
   end
 
+  get '/messages' do
+    @messages = Messages.messages
+    erb :'messages'
+  end
+
+  get '/add_message' do
```
EdwardAndress, line 19: The RESTful (conventional) route for returning the new peep form would be `GET "/peeps/new"`

EdwardAndress: or `GET "/messages/new"` if you prefer to call your class Message

EdwardAndress: There are also some conventions around what you call your class methods here, though they are less consistent between languages.  In Ruby, you would expect to see

Message.all
Message.create

Possibly because these are the method names used in the dominant framework, Rails.

lib/messages.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  attr_reader :messages, :date
+
+  def initialize(messages:, date:)
+    
+    @messages = messages
+    @date = date
+    
+  end
+
```
EdwardAndress, line 12: Nice that you're creating instances of Message here.

EdwardAndress:
Hi Micael, this is a good attempt and I think you've demonstrated the main learning objectives around reading from and writing to a database.  I would have liked to see a couple of unit tests, however – did you run out of time or did you get stuck?

The next step would probably be to add the ability to search Messages based on a keyword. Would you know how to get started with something like that? What would you plan be?

PR #175, opened 2022-01-24
Viktor - end result

ViktorVrecko:


app.rb
```diff
@@ -1,9 +1,35 @@
 require 'sinatra/base'
+require 'sinatra/reloader' # start auto reload
+require_relative './lib/chittermanager' 
 
 class Chitter < Sinatra::Base
+  register Sinatra::Reloader 
+
```
EdwardAndress, line 7: You can probably remove things like the `/test` route now that they're no longer needed.

app.rb
```diff
@@ -1,9 +1,35 @@
...
   get '/test' do
     'Test page'
   end
 
+  get '/' do
+    erb :index
+  end
+  
+  get '/view' do
```
EdwardAndress, line 16: Very nice that you're using the same route for both `.all` and `.filter`.  Not many people figure this one out.

lib/chittermanager.rb
```diff
@@ -0,0 +1,55 @@
...
+  
+  def self.create(peep: , time:)
+
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
+    
+    # result = connection.exec("INSERT INTO peeps (message, time) VALUES('#{peep}', current_timestamp) RETURNING id, message, time;")
```
EdwardAndress, line 33: You can probably remove this commented out code.

ViktorVrecko:
Thanks a lot for the review and the comments Eddie! I had a short eye on the those solutions shich handle the SQL part and will focus more on it by time.  I am very happy about the results :-)

PR #178, opened 2022-02-02
Susana

EdwardAndress:


EdwardAndress:
Hey Susana, I know from working with you that you've learned a lot since the time you did this :)

It's a good start though and I think you made the right call to begin by returning some hard coded values.  I've left a couple of specific comments against the code.

If you were to do it again now, what would you do differently?

app.rb
```diff
@@ -1,9 +1,30 @@
...
+    erb:'index'
+  end
+
+  get '/peep' do
+  peeps = 
+    [
+      "this is my first peep!",
+      "this is my second peep!",
+      "this is my third peep!"
+    ]
```
EdwardAndress, line 26: At the end of `get "/peep" do` you probably want to render a template like `erb :index`.  At the moment your test probably will pass but you'll just see the array on the screen, rather than a nice list of peeps

lib/peep.rb
```diff
@@ -0,0 +1,11 @@
+require 'pg'
+
```
EdwardAndress, line 2: The next step for this app would be to use your Peep class in app.rb... then think about adding the database

PR #180, opened 2022-02-07
 Kashif Atcha

siellsiell:
Hi Kashif,

I'm Simo, a coach. I've created a pull request for you so I can review your challenge.


siellsiell (commented):
It looks like you didn't get very far through the challenge although what I see looks promising! Did you run into any blockers?

PR #183, opened 2022-02-11
Chitter Challenge - Lynsey Barr

Lynsey-Barr:
# Your name

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,


app.rb
```diff
@@ -1,9 +1,19 @@
 require 'sinatra/base'
+require './lib/peep'
 
 class Chitter < Sinatra::Base
   get '/test' do
     'Test page'
   end
 
+  get '/messageboard' do
```
Katerina-codes, line 9: The naming could be better here, better naming of your route would be **get /peeps**

app.rb
```diff
@@ -1,9 +1,19 @@
...
+    @peep = Peep.all           
+    erb :messageboard
+  end
+
+  get '/chatter' do
```
Katerina-codes, line 14: The naming here is confusing, if we said that the route about for all peeps should be **get /peeps** then this should be **get /peeps/new** as we're creating a new peep.

spec/features/test_page_spec.rb
```diff
@@ -1,6 +1,20 @@
...
   end
+
+  scenario 'viewing the message board' do
+    visit('/messageboard')
+    expect(page).to have_content "Welcome to the Chitter Channel"
+    expect(page).to have_content "OMG loving these bangers!"
+  end
+
+  scenario 'opens chitter form to add peep' do
+    visit('/chatter')
```
Katerina-codes, line 16: This test isn't testing anything so it might be passing (falsely) but you need to be more specific and add an expectation.

PR #185, opened 2022-02-11
Chitter - Luke Collins

Lucifer-Kett:
# Your name

Luke/Luce Collins

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,8 +1,21 @@
 require 'sinatra/base'
+require 'sinatra/reloader'
+require './lib/peeps'
 
 class Chitter < Sinatra::Base
+  configure :development do
+    register Sinatra::Reloader
+  end
+
   get '/test' do
     'Test page'
+  end
+
+  get '/' do
+    ENV
+
+    @peeps = Peeps.all
+    erb(:index)
   end
 
   run! if app_file == $0
--- a/app.rb
+++ b/app.rb
@@ -18,6 +18,10 @@ class Chitter < Sinatra::Base
     erb(:index)
   end
 
+  get '/comment' do
+    erb(:comment)
+  end
+
   post '/' do
     Peeps.create(message = params[:message])
 
--- a/app.rb
+++ b/app.rb
@@ -18,5 +18,11 @@ class Chitter < Sinatra::Base
     erb(:index)
   end
 
+  post '/' do
+    Peeps.create(message = params[:message])
+
+    redirect '/'
+  end
+
   run! if app_file == $0
 end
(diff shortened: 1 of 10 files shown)


app.rb
```diff
@@ -1,9 +1,32 @@
 require 'sinatra/base'
+require 'sinatra/reloader'
+require './lib/peeps'
 
 class Chitter < Sinatra::Base
+  configure :development do
+    register Sinatra::Reloader
+  end
+
   get '/test' do
```
EdwardAndress, line 10: You can probably remove this bit :)

app.rb
```diff
@@ -1,9 +1,32 @@
...
+  get '/comment' do
+    erb(:comment)
+  end
+
+  post '/' do
+    Peeps.create(message = params[:message])
+
+    redirect '/'
+  end
+
```
EdwardAndress, line 30: You're almost there with the routing but it's not quite RESTful.

REST is a convention used for creating routes in a database backed web application, such as this. Each of your routes should contain a resource (pluralised) like `peeps` then, when combined with the HTTP verbs you can create unique, very predictable, routes (you also need to add `new` and `edit` in some cases).

For example...

```ruby
get '/peeps' do 
# for retrieving all the peeps
end

get '/peeps/new' do
# to get the form for creating a new peep
end

post '/peeps' do
# used when the new peep form is submitted
end

get '/peeps/edit' do
# used to get the form for editing a peep, if that is possible
end

put '/peeps/:id' do
# for when the edit peep form is submitted
# note that we also need the peep id here since we're going to update a specific peep
end
```

EdwardAndress: Using the above pattern, you can predict what the routes would be for any new resource, like `Post` or `Space` or `Event`... and so on

EdwardAndress: Deviating from this convention can lead to confusion.. for example when I saw this...

```ruby
get '/comment' do
    erb(:comment)
  end
```

I thought that it was possible to for users to add comments to a peep.

EdwardAndress:
Hi Luce, this is a good solution which, I think, demonstrates the main learning objectives. You can improve it by making the routes RESTful.  I've also added a comment about testing.

EdwardAndress:
The next step would be to add filtering. Do you know how you would approach that?

spec/features/all_feature_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'pg'
+
+feature 'shows chitter peeps' do
+  scenario '.all shows peeps' do
```
EdwardAndress, line 4: This is a feature test, so you're not testing the user journey, rather than a specific method.

PR #188, opened 2022-02-11
Barrington Brook- Chitter challenge

barringtonbrook:


app.rb
```diff
@@ -1,8 +1,32 @@
 require 'sinatra/base'
+require_relative './lib/peep'
 
 class Chitter < Sinatra::Base
-  get '/test' do
-    'Test page'
+
```
EdwardAndress, line 5: This is all pretty neat and it's great that you're thought to encapsulate some of the database logic in the `select_database` and `connect` methods.

The main things to reflect on in terms of improvements would be RESTful routing and the user journey.

`GET '/peeps'` and `POST '/peeps'` are both RESTful but `GET '/post'` is not, since you don't have a resource called `post`. The conventional route would be `GET '/peeps/new'` (for the route which returns the form to create a new `Peep`).

Regarding the user journey – at the moment, the user is redirected to the new peep form, from the `/` path, which may feel a bit odd. I'd probably expect to either see a welcome screen or the list of all peeps.  Interested to hear your thoughts on that though!

lib/peep.rb
```diff
@@ -0,0 +1,19 @@
+require 'pg' 
+
+class Peep
+ 
+  @@connection = nil
+
+  def self.connect(dbname)
+    @@connection = PG.connect(dbname: dbname)
+  end
+
```
EdwardAndress, line 10: Conventions will vary a little from language to language and framework to framework but, in Ruby, I'd expect these methods to be called `.all` for retrieving all Peeps and `.create` for creating a new Peep.

If you're not using Ruby in the long term, Ruby conventions are perhaps not important to you but there will be _some_ conventions in the language that you do use :)

spec/features/display_peeps_spec.rb
```diff
@@ -0,0 +1,6 @@
+feature 'display peeps' do
+  scenario 'navigate to page and list peeps from database' do
+    visit '/peeps'
+    expect(page).to have_content("This is a peep!")
+  end
```
EdwardAndress, line 5: I found it a bit confusing that there's nothing explicitly adding a peep to the DB here. I later found that you have a nice helper method which is called in `spec_helper`.

It's neat, but probably easier for people to follow along if you call that method here, in your test. You could also consider refactoring that helper method to take one arg, which would be the message content, so that it's really easy to see what's going on right from here.

EdwardAndress:
Hey Barrington! This is a really nice solution to the Chitter challenge. I particularly appreciate that you've put some thought into code quality.

There are a bunch of comments in line against the code, which are really focussed on helping you polish your approach to the challenge.  The one thing that I didn't mention there is that there are no unit tests. Did you run out of time or was there something blocking you there?

PR #189, opened 2022-02-11
Fred Halliwell

FredHalliwell:
Fred Halliwell 


spec/message_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+      connection.exec("INSERT INTO messages (name, message, time) VALUES ('fred', 'wow nice test', '#{Time.now}');")
+      connection.exec("INSERT INTO messages (name, message, time) VALUES ('phil', 'i love capybara', '#{Time.now}');")
+      connection.exec("INSERT INTO messages (name, message, time) VALUES ('jim', 'no SQL inject pls', '#{Time.now}');")
+
+      posts = Message.all
+
+      expect(posts[0]).to include 'jim'
+      expect(posts[1]).to include('i love capybara')
+      expect(posts[2]).to include('fred')
+    end
```
jconnollyweb, line 22: Great use of refactoring

app.rb
```diff
@@ -1,9 +1,40 @@
...
+    @new_message.post(@name, @message, @time)
+    @messages = Message.all
+    erb :index
+  end
+
+  post '/filtered' do
+    @filter = params[:filter]
+    @new_message = Message.new # is this wrong ????? does this make a whole new empty message ?
+    @messages = Message.filter(@filter)
+    erb :index
```
jconnollyweb, line 31: Good use of instances and refactoring.

app.rb
```diff
@@ -1,9 +1,40 @@
...
+class MessageManager < Sinatra::Base
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  get '/' do
+    erb :post_message
+  end
+
+  post '/messages' do
```
EdwardAndress, line 17: You only need to use instance vars here if you want those vars to be available in the view.

EdwardAndress: Nothing is broken by you using instance vars for things that are not needed in the view, but it's potentially confusing to the reader (and that's never good) :)

app.rb
```diff
@@ -1,9 +1,40 @@
...
+    @name = params[:name]
+    @message = params[:message]
+    @new_message = Message.new
+    @time = Time.now
+    @new_message.post(@name, @message, @time)
+    @messages = Message.all
+    erb :index
+  end
+
+  post '/filtered' do
```
EdwardAndress, line 27: You can actually use the same route for both filtered and unfiltered messages.  You would need to include an if statement, which checks to see if the user has sought to do some filtering, then you'd assign `@messages` accordingly, either filtered or unfiltered.  If the search term is submitted via a form and, is therefore in the params, you could do this...

```ruby
get '/posts' do
  if params['search']
    @messages = Message.filter(params['search']
  else
    @messages = Message.all
  end

  erb :index
end
```

lib/message.rb
```diff
@@ -0,0 +1,63 @@
...
+  def self.all
+    outputs = []
+    connection = if ENV['ENVIRONMENT'] == 'test'
+                   PG.connect(dbname: 'messages_app_test')
+                 else
+                   PG.connect(dbname: 'messages_app')
+                 end
+
+    result = connection.exec('SELECT * FROM messages ORDER BY user_id DESC;')
+    result.each do |message|
```
EdwardAndress, line 18: This is fine but it would be neater if you used `#map`

```ruby
result.map do |message|
  { name: message['name], message: message['message'], time: message['time'] }
end
```

Map always returns a new array, so this one line would replace 8.

EdwardAndress: OK three lines! :)

spec/feature/view_message_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    connection.exec("INSERT INTO messages (name, message, time) VALUES ('fred', 'wow nice test', '#{Time.now}');")
+    connection.exec("INSERT INTO messages (name, message, time) VALUES ('phil', 'i love capybara', '#{Time.now}');")
+    connection.exec("INSERT INTO messages (name, message, time) VALUES ('jim', 'dont SQL inject pls', '#{Time.now}');")
+
+    result = connection.exec('SELECT * FROM messages ORDER BY user_id DESC;')
+
+    visit('/messages')
+
+    expect(result[0]).to have_content 'jim'
+    expect(result[1]).to have_content 'i love capybara'
```
EdwardAndress, line 31: Glad to hear it

EdwardAndress:
Hi Fred! This is a nice solution which clearly demonstrates the learning objectives around reading from and writing to a database. Well done. I've left a few comments in line against your code.

PR #191, opened 2022-02-12
Cristina Gonzalez - Chitter

CristinaGonzalezVaro:
# Cristina Gonzalez

# User stories 

# README checklist

- [x] how to install,
- [x] how to run,


README.md
```diff
@@ -64,3 +62,10 @@ As a Maker
 So that I can find relevant peeps
 I want to filter on a specific keyword
 ```
+
+## Screenshots
+* Chitter homepage
+![alt text](https://github.com/CristinaGonzalezVaro/chitter-challenge-apprenticeships/blob/main/Chitter%20homepage.png)
+
+* Posting new peep
+![alt text](https://github.com/CristinaGonzalezVaro/chitter-challenge-apprenticeships/blob/main/Chitter%20new%20peep.png)
```
DWatts15, line 71: This seems like a really useful idea to include in your README!

README.md
```diff
@@ -1,12 +1,7 @@
 ## Chitter Challenge
 
-* Challenge time: until the end of the day
-* Feel free to use google, your notes, books etc but please work on your own
-* Please raise a pull request when you start this challenge, and keep pushing updates as and when you make commits throughout the day
-* There is _no expectation_ to finish all or any of the user stories, please use this time to reflect on where you feel you are with the skill and what may support your learning.
-* If you get blocked, please reflect on what blocked you and any strategies you adopted that helped you make progress.
+Welcome to Chitter ~~not a clone of Twitter~~
```
DWatts15, line 3: :)

lib/peep.rb
```diff
@@ -0,0 +1,25 @@
+require 'pg'
+
+class Peep
```
DWatts15, line 3: Calling your class 'Peep' was a good idea. I have called mine 'Message' and I think this may cause confusion later as I have also used 'message' as a parameter and variable.

lib/peep.rb
```diff
@@ -0,0 +1,25 @@
...
+  def self.all
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
+
+    result = connection.exec("SELECT * FROM peeps;")
```
DWatts15, line 11: A minor improvement here may be to use #exec_params to secure the app against SQL injection (like we saw in Challenge 12 of last week's pair programming challenge)

spec/peep_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+      
+    end
+  end
+
+  describe '.create' do
+    it 'posts a new peep' do
+      Peep.create(message: "Thinking of something funny to peep")
+      expect(Peep.all).to include "Thinking of something funny to peep"
+    end
+  end
```
DWatts15, line 20: I think this test should also check that the created Peep has been added to the database by querying the database (eg. SELECT * FROM peeps...)

views/peeps/index.erb
```diff
@@ -0,0 +1,46 @@
...
+  <body>
+    <h1 style="text-align:center;"> Welcome to Chitter </h1>
+    <h2 style="text-align:center;"> See what's happening </h2>
+
+    <h3 style="text-align:center;"><a href="/peeps/new"> Peep Something </a></h3>
+
+    <ul>
+      <% @peeps.each do |peep| %>
+        <li><%= peep %></li>
+        <br>
```
DWatts15, line 39: This line break(?) is a simple but really good idea to keep things tidy on the page - I should have done this!

DWatts15 (commented):
Your code is very easy to follow as class, variables and methods are all named clearly. I like that you have thought about your code from the user's perspective - the css design in the erb pages is pleasing to the eye, and I like your subheadings and buttons ("See what's happening" and "Peep something") as they are far more interesting than just "Submit". 
I think one improvement to make is to double-check your test for .create as it probably needs to check that the new Peep has been added to the database within this test - I can see that you are checking the database by calling Peep.all, but I think this is the kind of coupling we are encouraged to avoid (although I'm not 100% sure so you might want to check that!).

README.md
tomfuller: Agreed! The screenshots are a really nice touch

lib/peep.rb
```diff
@@ -0,0 +1,25 @@
+require 'pg'
+
+class Peep
+  def self.all
+    if ENV['ENVIRONMENT'] == 'test'
```
tomfuller, line 5: This if statement is repeated later in the file, is there a refactor you could do to make this file more DRY?

tomfuller:
This is a really nice solution, you endpoints are RESTful and your README is great! I have a little note around some code duplication and also some advice around mocking the database connection in your unit tests. In unit tests we only want to test a single unit of code and ideally mock out anything else. This is so we know where the problem lies if there is one and also it reduces the cost(runtime) of the unit tests.

spec/peep_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'peep'
+
+describe Peep do
+
+  describe '.all' do
+    it 'shows all peeps' do
```
tomfuller, line 6: Its really great that you are doing unit tests as well as your feature tests! My only thought is that for unit tests it might be better to mock the database and only test the logic in your function. For example if you mocked the database to return a peep and then you can check that your map function works and returns the message part.

PR #194, opened 2022-02-13
Chitter Challenge

benfielding74:
# Ben FIelding

# User stories 

- [ x ] User story 1: "I want to see all the messages (peeps) in a browser"
- [ x ] User story 2: "I want to post a message (peep) to chitter"
- [ x ] User story 3: "I want to see the date the message was posted"
- [ x ] User story 4: "I want to see a list of peeps in reverse chronological order"

# README checklist

- [ x ] how to install,
- [ x ] how to run,
- [ x ] and how to test your code?


docs/README.md
```diff
@@ -0,0 +1,66 @@
...
+```
+As a Maker
+So that I can easily see the latest peeps
+I want to see a list of peeps in reverse chronological order
+```
+```
+As a Maker
+So that I can find relevant peeps
+I want to filter on a specific keyword
+```
```
shalini9, line 66: Great artwork for posting peeps

lib/chitter_manager.rb
```diff
@@ -0,0 +1,38 @@
...
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else 
+      connection = PG.connect(dbname: 'chitter')
+    end
+
+    result = connection.exec("SELECT * FROM peeps ORDER BY peep_date DESC;")
+    result.map do |peep| 
+      ChitterManager.new(id: peep['id'], message: peep['message'], date: peep['peep_date'])
+    end
```
shalini9, line 23: easy to read code

lib/chitter_manager.rb
```diff
@@ -0,0 +1,38 @@
...
+  def self.all
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else 
+      connection = PG.connect(dbname: 'chitter')
+    end
+
+    result = connection.exec("SELECT * FROM peeps ORDER BY peep_date DESC;")
+    result.map do |peep| 
+      ChitterManager.new(id: peep['id'], message: peep['message'], date: peep['peep_date'])
```
shalini9, line 22: very good wrapping objects

README.md
```diff
@@ -52,15 +27,51 @@ As a Maker
...
+* Create the database using the `psql` command `CREATE DATABASE chitter_test`;
+* Connect to the database using the `psql` command `\c chitter_test;`
+* Run the query we have saved in the file `01_create_chitter_table.sql`
+
+## To Run
+
+* `git clone https://github.com/benfielding74/chitter-challenge-apprenticeships.git`
+* `bundle install`
+* `ruby app.rb`
+* `visit "https://localhost:4567"`
```
siellsiell, line 77: Nice and comprehensive README :)

app.rb
```diff
@@ -1,8 +1,20 @@
...
+require './lib/chitter_manager'
 
 class Chitter < Sinatra::Base
-  get '/test' do
-    'Test page'
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  get '/chitter' do
```
siellsiell: A simpler solution could be to use the index route `/` for the landing page. That also matches the fact that you're using a template called `index` at line 12.

lib/chitter_manager.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+  end
+
+  def self.post(message:)
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else 
+      connection = PG.connect(dbname: 'chitter')
+    end
```
siellsiell, line 31: You're using this same piece of code in more than one place in this file. How might you get rid of this duplication?

lib/chitter_manager.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    peep = message.gsub(/'/, "''")
```
siellsiell, line 33: Interesting. What caused you to add this?

spec/chitter_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+describe ChitterManager do
+
+  describe '.all' do
+    it 'returns a list of all messages' do
+      connection = PG.connect(dbname: 'chitter_test')
+      ChitterManager.post(message: 'This is my first peep')
+
+      peeps = ChitterManager.all
+
+      expect(peeps[0]).to be_a ChitterManager
```
siellsiell, line 12: To make the test stronger, you could also check that the array only contains one item.

lib/chitter_manager.rb
siellsiell, line 22: It's a bit strange that `ChitterManager` is also used to wrap the data for a single peep. I would expect a separate class for that, e.g. `Peep`, or instead, that `ChitterManager` is renamed to `Peep`. 

In software, calling a class `Manager` is often taken to imply that it manages access or controls a bunch of functionality (and your class does this well with the `all` and `post` methods). This means that implicitly we expect there to only ever be one manager, not one manager per peep.

siellsiell (commented):
Hey it's Simo, a coach :)

Thanks for submitting your challenge code and for filling out the pull request description! This was a good attempt at the challenge. Your code is very readable, well documented and well tested. One thing that's missing is the migration file. Since you added a date column to the Peeps table, there should be a migration to reflect that.

I left a few more detailed comments for you to have a look at. Feel free to ask any if you have questions!

README.md
benfielding74: Thanks Simo

app.rb
benfielding74: I have updated this to make it simpler. Thanks

lib/chitter_manager.rb
benfielding74: When I was testing I realised that peeps such as "It's Saturday" or "They're here" would escape my database insertion command and throw errors. I was familiar with regex from some javascript tutorials I had done last year so I looked for a Ruby solution and found gsub to insert another single quote and stop it escaping when going into the database.

lib/chitter_manager.rb
benfielding74: Thanks I will refactor this to make it more coherent.

PR #195, opened 2022-02-14
chitter-challenge

sapana2031:
# Your name
Sapna Tavrawala

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,9 +1,25 @@
 require 'sinatra/base'
+require 'sinatra/reloader'
+require './lib/chitter'
 
 class Chitter < Sinatra::Base
+  configure :development do
+    register Sinatra::Reloader
+  end
+  
   get '/test' do
     'Test page'
   end
 
+  get '/message' do
+    @messages = ChitterMessage.all
+    erb :index
+  end
+
+
+
+  
+
+  
   run! if app_file == $0
 end
--- a/app.rb
+++ b/app.rb
@@ -21,9 +21,7 @@ class Chitter < Sinatra::Base
   end
 
   post '/newpost' do
-    message = params['message']
-  connection = PG.connect(dbname: 'chitter')
-  connection.exec("INSERT INTO peeps (message) VALUES('#{message}')")
+    ChitterMessage.create(message: params[:message])
   redirect '/message'
   end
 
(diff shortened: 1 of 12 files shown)


app.rb
```diff
@@ -1,9 +1,34 @@
...
+  get '/newpost' do
+    erb :new
+  end
+
+  post '/newpost' do
+    ChitterMessage.create(message: params[:message])
+  redirect '/message'
+  end
+
+
```
laurenbethgray, line 28: Really clean, clear code that is easy to read and understand.

spec/features/posting_messages_spec.rb
```diff
@@ -0,0 +1,10 @@
+feature 'Posting new message with date' do
+  scenario 'A user can post a message to ChitterMessage' do
+    visit('/newpost')
+    fill_in('message', with: 'I am fine.')
+    click_button('Submit')
+
+    expect(page).to have_content 'I am fine.'
+    #expect(page).to have_content Date.today
```
laurenbethgray, line 8: Remember to remove commented out code for final commit.

EdwardAndress:
Hi Sapna! I think you've done a good job of demonstrating the main learning objectives here.  There are one or two places where you could improve the app though.

**RESTful routing**
At the moment, your app is not following the REST convention.  This convention exists to make it easier for people to understand a codebase, when they move from one project to another, and to use APIs (which should also use RESTful routing). The idea is that by combining the name of the resource and the HTTP method, you can (almost) create all the routes you need.  There are some exceptions.  See below.

To get a list of all messages
request: `GET "/chitter_messages"`
controller: `get "/chitter_messages" do`

To get the form for creating a new message
request: `GET "/chitter_messages/new"`
controller: `get "/chitter_messages/new" do`

To create a new message (submit the form)
request: `POST "/chitter_messages"`
controller: `post "/chitter_messages" do`

In every case, the name of the resource is in the url as `chitter_messages` and we combine it with different HTTP methods (or add `/new`) for each request / controller action.

EdwardAndress:
The next step would be to allow filtering of the messages. Do you have an idea as to how that could be done?

PR #198, opened 2022-02-14
Sonali Kar- user story 1,2 & 3 completed

sonalikr:
# Your name

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -13,8 +13,18 @@ class Chitter < Sinatra::Base
 
   get '/messages' do
     @messages = ChitterMessage.all
-    # @messages = ["This is a peep!", "This is an emergency!"]
     erb :index
+  end
+
+  get '/new' do
+    erb :new
+  end
+
+  post '/new' do
+    message = params['message']
+    connection = PG.connect(dbname: 'chitter')
+    connection.exec("INSERT INTO peeps (message) VALUES('#{message}')")
+    redirect '/messages'
   end
 
   run! if app_file == $0
--- a/app.rb
+++ b/app.rb
@@ -22,8 +22,10 @@ class Chitter < Sinatra::Base
 
   post '/new' do
     message = params['message']
+    messagedate = params['messagedate']
+    
     connection = PG.connect(dbname: 'chitter')
-    connection.exec("INSERT INTO peeps (message) VALUES('#{message}')")
+    connection.exec("INSERT INTO peeps (message, date) VALUES('#{message}', '#{messagedate}')")
     redirect '/messages'
   end
 
--- a/app.rb
+++ b/app.rb
@@ -1,8 +1,20 @@
 require 'sinatra/base'
+require 'sinatra/reloader'
+require './lib/chitter'
 
 class Chitter < Sinatra::Base
+  configure :development do
+    register Sinatra::Reloader
+  end
+
   get '/test' do
     'Test page'
+  end
+
+  get '/messages' do
+    @messages = ChitterMessage.all
+    # @messages = ["This is a peep!", "This is an emergency!"]
+    erb :index
   end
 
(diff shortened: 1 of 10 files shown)


spec/chitter_spec.rb
```diff
@@ -2,9 +2,8 @@
 
 describe '.all' do
   it 'returns the messages of peeps' do
```
MakersKyrstenStone, line 4: could you explain why you need two separate variables that equal the sam e thing here please?

lib/chitter.rb
```diff
@@ -2,10 +2,17 @@
...
 
+ 
+
   def self.all
+    objarr = []
     connection = PG.connect(dbname: 'chitter')
     result = connection.exec('SELECT * FROM peeps')
-    result.map { |message| message['message'] }
+    result.each do |row|
+      objarr.push({"msg" => row["message"], "dt" => row["date"]})
```
MakersKyrstenStone, line 12: is this line putting all the elements in the returned array in 'result' into a new hash?

app.rb
```diff
@@ -22,8 +22,10 @@ class Chitter < Sinatra::Base
 
   post '/new' do
     message = params['message']
+    messagedate = params['messagedate']
+    
     connection = PG.connect(dbname: 'chitter')
-    connection.exec("INSERT INTO peeps (message) VALUES('#{message}')")
+    connection.exec("INSERT INTO peeps (message, date) VALUES('#{message}', '#{messagedate}')")
```
MakersKyrstenStone, line 28: are both of these connection.exec statements needed? does the second one not complete the same task as the first one?

views/index.erb
```diff
@@ -1,5 +1,5 @@
 <ul>
-  <% @messages.each do |message| %>
-    <li><%= message %></li>
+  <% @messages.each do |msg_content| %>
+    <li><%= msg_content['msg'] %> <%= msg_content['dt'] %></li>
```
MakersKyrstenStone, line 3: do you need both sets of <% %> in here as the colours don't seem to match up?

db/migrations/02_add_date_to_chittertable.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE peeps ADD COLUMN date VARCHAR(60);
```
MakersKyrstenStone, line 1: this is great, i've not seen this before

MakersKyrstenStone (commented):
I really enjoyed reading this it made sense and was very nicely laid out, as some bits were more advanced than i had got i had a few questions which i noted but from my point of view it looks like you did great Sonali.

app.rb
```diff
@@ -1,9 +1,33 @@
...
   get '/test' do
     'Test page'
   end
 
+  get '/messages' do
+    @messages = ChitterMessage.all
+    erb :index
+  end
+
+  get '/new' do
```
Katerina-codes, line 19: This route name needs to be more explicit as it is not clear what new is referring to. Instead, **get /messages/new** would be much clearer.

app.rb
```diff
@@ -1,9 +1,33 @@
...
+    erb :new
+  end
+
+  post '/new' do
```
Katerina-codes, line 23: Same issue as above here with naming, **post /messages/new** would be better.

app.rb
```diff
@@ -1,9 +1,33 @@
 require 'sinatra/base'
+require 'sinatra/reloader'
+require './lib/chitter'
 
 class Chitter < Sinatra::Base
+  configure :development do
+    register Sinatra::Reloader
+  end
+
   get '/test' do
```
Katerina-codes, line 10: This test page doesn't need to be here so it can be deleted :)

PR #202, opened 2022-02-14
Chitter Challenge - Jaabir Yusuf

JaabirM:
# Your name

Jaabir Yusuf

# User stories 

- [ X ] User story 1: "I want to see all the messages (peeps) in a browser"
- [ X ] User story 2: "I want to post a message (peep) to chitter"
- [ X ] User story 3: "I want to see the date the message was posted"
- [ X ] User story 4: "I want to see a list of peeps in reverse chronological order"
- [ X ] User story 5: "I want to filter on a specific keyword"

# README checklist

- [ X ] how to install,
- [ X ] how to run,


EdwardAndress:
Hello Jaabir! This is a nice solution which clearly demonstrates the learning objectives around reading from and writing to a database.  So well done there.

It's also a very interesting solution in terms of how you've approached the challenge.  I think there are a few places where this could be polished a little though.

The first thing would be to add some tests. Is this something you've struggled with or did you prioritise features this time?  I'm happy to provide some guidance if it's the former.

The other thing to reflect on is SRP.  At the moment, you have a single class `Chitter_message` which interacts with the database _and_ generates HTML – so it has this additional responsibility that is to do with presentation. If you want to abstract this process of generating HTML away from your template, you could create a MessagePresenter class for that purpose. Such classes are used sometime in Rails applications. So you could probably find some examples there.

PR #203, opened 2022-02-14
Chitter challenge Shalini Nangia

shalini9:
# Your name

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,9 +1,19 @@
 require 'sinatra/base'
+require 'sinatra/reloader'
 
 class Chitter < Sinatra::Base
+  configure :development do
+    register Sinatra::Reloader
+  end
+
   get '/test' do
     'Test page'
   end
+
+  get '/peeps' do
+    @peeps = Peep.all
+    erb :'peeps/index'
+  end 
 
   run! if app_file == $0
 end
--- a/spec/peep_spec.rb
+++ b/spec/peep_spec.rb
@@ -1,14 +1,17 @@
 require 'peep'
 
-describe Peep do
-  describe '.all' do
-    it 'returns all peeps' do
-      peeps = Peep.all
+describe '.all' do
+  it 'returns a list of peeps' do
+    add_row_to_test_database
 
-      expect(peeps).to include("This is a peep!")
-      expect(peeps).to include("Meditate to evolve.")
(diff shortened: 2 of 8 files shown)


app.rb
```diff
@@ -1,9 +1,19 @@
 require 'sinatra/base'
+require 'sinatra/reloader'
 
 class Chitter < Sinatra::Base
+  configure :development do
+    register Sinatra::Reloader
+  end
+
   get '/test' do
```
EdwardAndress, line 9: You can probably remove this :)

spec/peep_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+    peeps = Peep.all
+    expect(peeps).to include('This is a peep!')
+    expect(peeps).to include('Meditate to evolve.')
+    expect(peeps).to include('Stay in the present moment.')
+    expect(peeps).to include('Being of service to others is what brings true happiness.')
+  end
+end
+
+
+
```
EdwardAndress, line 16: This test, and the feature test, are both great. I would be tempted to do a little refactor though because the `add_row_to_database` method sounds like it's going to add just one row but it actually adds 4 and, also, from the test you cannot see what the text will be.

You could consider using a helper method like this instead

```ruby
def add_row_to_database(message)
  connection.exec("INSERT INTO peeps (message) values (#{message});")
end
```

Then you could use it in your test like this...

```ruby
describe '.all' do
  it 'returns a list of peeps' do
    add_row_to_test_database('This is a peep!')
    add_row_to_test_database('Meditate to evolve.')
    add_row_to_test_database('Stay in the present moment.')
    add_row_to_test_database('Being of service to others is what brings true happiness.')

    peeps = Peep.all
    expect(peeps).to include('This is a peep!')
    expect(peeps).to include('Meditate to evolve.')
    expect(peeps).to include('Stay in the present moment.')
    expect(peeps).to include('Being of service to others is what brings true happiness.')
  end
end
```

And this means you can see what is happening in the test a bit more easily.

EdwardAndress:
Hey Shalini, you've made a good start here and have partly demonstrated the main learning objectives for the week.  The next step would be to allow the user to create a Peep. Did you run out of time or did you get stuck?

PR #204, opened 2022-02-14
Kyrsten end of friday user story 2

MakersKyrstenStone:
# Your name

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,


app.rb
```diff
@@ -1,9 +1,27 @@
 require 'sinatra/base'
+require './lib/messages'
 
 class Chitter < Sinatra::Base
+ 
   get '/test' do
     'Test page'
   end
```
siellsiell, line 8: Its good practice to remove debug/test code when submitting code (anywhere, not just at Makers). It saves a bunch of annoying comments from teammates that point them all out :)

app.rb
```diff
@@ -1,9 +1,27 @@
...
+  get '/view_messages' do
+    @messages = Messages.all
+    erb :'messages/index'
+  end
+  
+  get '/message/new' do
+    erb :"messages/new"
+  end
+  post '/new_message' do
+    p "message received"
```
siellsiell, line 19: Same here.

spec/message_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+      # connect to database
+      connection = PG.connect(dbname: 'chitter_test')
+      # add some data for testing
+      connection.exec("INSERT INTO peeps (message) VALUES('This is a peep!');")
+      connection.exec("INSERT INTO peeps (message) VALUES('and another peep');")
+
+      messages = Messages.all
+
+      expect(messages).to include("This is a peep!")
+      expect(messages).to include("and another peep")
```
siellsiell, line 15: You could make this stronger by using `eq` instead of `include`. With `eq`, you could compare the whole array at once and on top of that, you can make sure it doesn't contain anything _other_ than what you expect.

spec/spec_helper.rb
```diff
@@ -41,6 +41,7 @@
 RSpec.configure do |config|
   config.before(:each) do
     setup_test_database
+    add_row_to_test_database
```
siellsiell, line 44: nice way to keep your code DRY :)

siellsiell (commented):
Thanks for submitting your challenge code! This was a good implementation of the first two user stories. Thanks for sharing your diagram too. It would be a good addition to a more detailed README.

I left more detailed comments below. You can use the reply feature on each comment if you have questions :)

PR #206, opened 2022-04-01
Chitter Challenge

Addeak:
# Your name

Adam Deakin

# User stories 

# README checklist

- [x] how to install,
- [x] how to run,


app.rb
```diff
@@ -1,9 +1,34 @@
...
+
+  enable :sessions, :method_override
+
+  configure :development do
+    register Sinatra::Reloader
+  end
+
   get '/test' do
     'Test page'
   end
```
alicelieutier, line 16: You shold probably remove this :)

database_connection_setup.rb
```diff
@@ -0,0 +1,7 @@
+require './lib/database_connection'
+
+if ENV['ENVIRONMENT'] == 'test'
+  DatabaseConnection.setup('chitter_test')
+else
+  DatabaseConnection.setup('chitter')
+end
```
alicelieutier, line 7: This would probably be better inside a method, so you can call it at specific points in the code, rather than have it run everytime the file is required.

lib/database_connection.rb
```diff
@@ -0,0 +1,13 @@
+require 'pg'
```
alicelieutier, line 1: Actually, I'm thinking the file for connection setup should probably be merged with this one. It would keep things together.

alicelieutier: I guess not merging mean keeping this more "generic". But it does create more files.

lib/peep.rb
```diff
@@ -0,0 +1,45 @@
...
+      Peep.new(
+        id: peep['id'],
+        message: peep['message'],
+        date_posted: peep['date_posted']  
+      )
+    end
+  end
+
+  def self.filter(keyword:)
+    result = DatabaseConnection.query("SELECT * FROM peeps WHERE message LIKE '%#{keyword}%';")
```
alicelieutier, line 36: Why do you use this syntax here instead of params like above for create? Not saying it's wrong, I'm just curious if there's a reason.
In general I would try to be consistent in syntax across a whole codebase.

spec/database_connection_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'database_connection'
+
+describe DatabaseConnection do
+  describe '.setup' do
+    it 'establishes a connection to a database via PG' do
+      expect(PG).to receive(:connect).with(dbname: 'chitter_test')
+
+      DatabaseConnection.setup('chitter_test')
```
alicelieutier, line 8: So, typically, what we do is set an environment variable in our spec_helper, and that will be picked up by the database connection setup to connect to the right db.
Check the model spec_helper here: https://github.com/makersacademy/course/blob/main/pills/ruby_web_project_setup_list.md#4-make-capybara-talk-to-sinatra

Otherwise, you whole:
```rb
if ENV['ENVIRONMENT'] == 'test'
  DatabaseConnection.setup('chitter_test')
else
  DatabaseConnection.setup('chitter')
end
```
Is not really useful.

By doing this here, you rely on the fact that this line executes after the line in app.rb.

spec/features/filtering_by_keyword_spec.rb
```diff
@@ -0,0 +1,19 @@
+feature 'Filtering by keyword' do
+  scenario 'User enters a keyword and a list of peeps containing keyword is shown' do
```
alicelieutier, line 2: This is a great feature test!

spec/features/reverse_order_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+
+    fill_in 'message', with: 'Oldest sibling'
+    click_button 'Submit'
+    fill_in 'message', with: 'Middle sibling'
+    click_button 'Submit'
+    fill_in 'message', with: 'Youngest sibling'
+    click_button 'Submit'
+
+    expect(first('.peep')).to have_content 'Youngest sibling'
+    expect(all('.peep').last).to have_content 'Oldest sibling'
```
alicelieutier, line 13: That's a nice way to do it.
Another way you can use is a regular expression on the content. This way, no need to refer to the classes:
```rb
expect(page).to match(/Youngest sibling.*Middle sibling.*Oldest sibling/)
```

spec/peep_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+require 'database_helpers'
+
+describe Peep do
+  describe '.create' do
+    it 'creates a new peep' do
+      peep = Peep.create(message: 'This is a peep!')
+      persisted = persisted_data(id: peep.id)
+      
+      expect(peep).to be_a Peep
+      expect(peep.id).to eq persisted['id']
```
alicelieutier, line 11: I think I'd check that the persisted data message match, rather than checking the id match, given that the id is what you set on line 8.

spec/peep_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+      Peep.create(message: 'And this is another one!')
+      Peep.create(message: 'Woah! And this is yet another!')
+
+      peeps = Peep.all
+
+      expect(peeps.length).to eq 3
+      expect(peeps.first).to be_a Peep
+      expect(peeps.first.id).to eq peep.id
+      expect(peeps.first.message).to eq 'This is a peep!'
+      expect(peeps.first.date_posted).to eq "#{Date.today}"
```
alicelieutier, line 28: Whenever you test for time, it's always better to stub the time than to rely on "today". Use a gem like 'Timecop' to freeze the time on a given date.

alicelieutier (commented):
Really good, well organised a clear.
I added some comments below, let me know if you have any questions.

PR #207, opened 2022-04-01
set up directory

khushnoodasif:
# Your name

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,


app.rb
```diff
@@ -1,9 +1,37 @@
 require 'sinatra/base'
+require_relative './lib/message_handler'
 
 class Chitter < Sinatra::Base
   get '/test' do
     'Test page'
   end
```
siellsiell, line 7: It's always good to remember to remove test code before submitting code for review (anywhere, not just at Makers)

app.rb
```diff
@@ -1,9 +1,37 @@
...
 
+  get '/message_page' do
```
siellsiell, line 9: Typically, I would expect the messages to be shown on the index page (the `/` route) of the app as this is where the user lands by default. If you want to keep the `/message_page` route, you should have a redirect that will take the user from`/` to `/message_page`.

lib/database_handler.rb
```diff
@@ -0,0 +1,26 @@
+require 'pg'
+
+class Database
+  def self.connect 
+    return PG.connect(dbname: 'chitter_test') if ENV['RACK_ENV'] == 'test'  
+    return PG.connect(dbname: 'chitter')
+  end 
+
+  def self.all
+    connection = Database.connect
```
siellsiell, line 10: Nice use of a method to reduce duplication in your code!

lib/database_handler.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def self.all_filtered(filter:)
+    connection = Database.connect
+    result = connection.exec("SELECT * FROM peeps WHERE messages LIKE '%#{filter}%'")
+  end 
+
+  def self.add(message:)
+    connection = Database.connect
+    time = Time.now
+    time = time.strftime("%d/%m/%Y")
```
siellsiell, line 22: Storing time as a string works but as you can see, it requires you to explicitly format it before putting it in the database. That ties your View to your Model -- ideally you would store all values in a format that isn't tied to how you want to present it to the user so that you can easily change the latter if needed without changing your database tables. For example, what if later there was a requirement that dates shown to the user should also include time information, not just the date? Your database would have no timing information for old messages.

Luckily, most databases actually provide a datatype to store time information: https://www.postgresql.org/docs/9.1/datatype-datetime.html

db/migrations/01_create_chitter_table.sql
```diff
@@ -1 +1 @@
-CREATE TABLE peeps(id SERIAL PRIMARY KEY, message VARCHAR(60));
+CREATE TABLE peeps(id SERIAL PRIMARY KEY, message VARCHAR(60));
```
siellsiell, line 1: Nice migration! You're missing one for adding the timestamp though.

lib/database_handler.rb
```diff
@@ -0,0 +1,26 @@
...
+    result = connection.exec('INSERT INTO peeps (messages, timestamp) VALUES($1,$2)',
```
siellsiell, line 23: Minor: `timestamp` isn't the best name for that column as currently you're using it to store only a date. Typically in programming, when we say timestamp, we mean a Unix timestamp, which is a number in seconds: https://www.unixtimestamp.com/

spec/features/message_page_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    visit('/filter')
+    fill_in('filter', with: 'Hello')
+    click_on 'Filter Messages'
+    expect(page).to have_content 'Hello World!'
+    expect(page).to_not have_content 'Bye World!'
+    click_on 'Reset Filter'
+    expect(page).to have_content 'Hello World!'
+    expect(page).to have_content 'Bye World!'
+  end
+end
```
siellsiell, line 47: Really thorough feature tests, nice!

spec/message_handler_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'pg'
+require_relative '../lib/message_handler'
+
+describe MessageHandler do
+  describe '.all' do 
+    let(:db_entry) { ["id" => "1047", "messages" => "Hello World", "timestamp" => "01/04/2022"] }
+    let(:db_double) { double(:dbdouble, add: db_entry, all: db_entry) }
```
siellsiell, line 7: I like how you've made testing with mocks possible by having a separate database class.

spec/database_handler_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    it 'connection to database' do 
+      expect(Database.connect).to be_a_instance_of(PG::Connection)
+    end 
+  end
+
+  describe '.all' do
+    it 'returns the whole database' do 
+      Database.add(message: "hello")
+      expect(Database.all.first['messages']).to eq 'hello'
+    end
```
siellsiell, line 15: It might be worth testing that `Database.all` also works when there is more than one message in the db

siellsiell (commented):
Hey, I'm a coach. Thanks for submitting!

This is a good solution. Well done :) See more details in the comments.

PR #214, opened 2022-04-01
Chitter PR

lacurnow:
# Your name
Lucinda Curnow

# User stories 

- [X] User story 1: "I want to see all the messages (peeps) in a browser"
- [X] User story 2: "I want to post a message (peep) to chitter"
- [X] User story 3: "I want to see the date the message was posted"

# README checklist

- [ ] how to install,
- [ ] how to run,

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -13,7 +13,6 @@ class Chitter < Sinatra::Base
   end
 
   post '/view_peeps/add' do
-    
     #PeepViewer.add(message: params['new_peep'])
     redirect '/view_peeps'
   end
--- a/app.rb
+++ b/app.rb
@@ -12,5 +12,11 @@ class Chitter < Sinatra::Base
     erb :view_peeps
   end
 
+  post '/view_peeps/add' do
+    
+    #PeepViewer.add(message: params['new_peep'])
+    redirect '/view_peeps'
+  end
+
   run! if app_file == $0
 end
--- a/app.rb
+++ b/app.rb
@@ -13,7 +13,7 @@ class Chitter < Sinatra::Base
   end
 
   post '/view_peeps/add' do
-    #PeepViewer.add(message: params['new_peep'])
+    PeepViewer.add(message: params['message'])
     redirect '/view_peeps'
   end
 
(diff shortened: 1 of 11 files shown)


khushnoodasif:
From what I have observed this code accomplishes what is supposed to do.

khushnoodasif:
- [x]  Was the code easy to understand? Yes
- [x]  Can the readability of the code be improved by smaller methods? Methods used were simplified to a great extent.
- [x]  Can the readability of the code be improved by different function/method or variable names? Readability is great but I would suggest using rubocop to see if there is anything with code that could be improved such as adding a newline after the end of code for better readability.
- [x]  Is the code located in the right file/folder/package? Yes

app.rb
```diff
@@ -1,9 +1,22 @@
-require 'sinatra/base'
+require 'sinatra'
+require 'sinatra/reloader' if development?
+require './lib/peepviewer'
+require './lib/peep'
 
 class Chitter < Sinatra::Base
```
EdwardAndress, line 6: You could probably remove the route for the test page

app.rb
```diff
@@ -1,9 +1,22 @@
...
   get '/test' do
     'Test page'
   end
 
+  get '/view_peeps' do
```
EdwardAndress, line 11: This is great but the naming is a little unconventional.

I'd expect the routes to be

`GET '/peeps'` for viewing all peeps
`POST '/peeps'` for creating a new peep

And I would expect the model to be called `Peep` so in the controller you'd do things like `Peep.all` and `Peep.create`.

There are also conventions for template names :)

`index` for the page that shows all the peeps

EdwardAndress: Ah! I see that there _is_ a model called `Peep` and that you've implemented another class which takes care of the database interaction. This is a nice idea! I'd probably call that class something like `PeepRepository`

EdwardAndress:
This is a nice solution! I particularly appreciate the fact that you've thought about code quality – this is reflected in your decision to implement the `PeepViewer` class. You've clearly demonstrated the learning objectives for this module.

PR #215, opened 2022-04-01
1,2,3,4 user stories done

agagotowiec:
# Agnieszka Gotowiec

# User stories 

# README checklist

- [x] how to install,
- [x] how to run,


mohamedadan92 (commented):
Great layout and easy to read app. No errors noticable

app.rb
```diff
@@ -1,8 +1,24 @@
 require 'sinatra/base'
+require './lib/post'
 
 class Chitter < Sinatra::Base
-  get '/test' do
-    'Test page'
+  get '/' do
+    'Chitter App'
```
alicelieutier, line 6: detail - remove unused line

app.rb
```diff
@@ -1,8 +1,24 @@
 require 'sinatra/base'
+require './lib/post'
 
 class Chitter < Sinatra::Base
```
alicelieutier, line 4: Nice use of RESTful routes!

lib/post.rb
```diff
@@ -0,0 +1,35 @@
...
+    @author = author
+    @message = message
+  end
+
+  def self.all
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
```
alicelieutier, line 18: These lines are present in all your methods that access the db - they could be extracted into a database helper class :)

lib/post.rb
```diff
@@ -0,0 +1,35 @@
...
+    result = connection.exec("SELECT * FROM peeps ORDER BY date DESC")
```
alicelieutier, line 19: I'm surprised queries without `;` work. I guess the framework adds the `;`. I would still make sure to add it however.

lib/post.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def self.create(date:, author:, message:)
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
+
+    result = connection.exec_params("INSERT INTO peeps (date, author, message) VALUES('#{date}', $1, $2) RETURNING id, date, author, message;", [author, message])
```
alicelieutier, line 32: Nice that you are [using exec_params to prevent sql injections](https://github.com/makersacademy/course/blob/main/apprenticeships_bookmark_manager/walkthroughs/12.md) here, but I'm wondering why you are not also doing that for the date, especially since the date is sent by the front-end as a param.

It would be different if the date was automatically added in ruby to peeps when they are created, and not coming from the front-end.

I know you form enforces a format for the input, but a malicious user could easily change the front-end html and send data in a different way, so your server should always protect itself from the front-end.

alicelieutier (commented):
Clear code, nicely separated and organised, well done!
I added a few comments below, let me know if anything doesn't make sense.

spec/features/viewing_posts_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  scenario 'visiting /posts shows new messages first' do
+    
+    Post.create(date: "2022-04-01", author: "Kate", message: "How are you?")
+    Post.create(date: "2022-06-01", author: "Kate", message: "How are you?")
+    Post.create(date: "2022-08-01", author: "Kate", message: "How are you?")
+
+    visit('/posts')
+    expect(page).to have_content("2022-08-01")
+    expect(page).to have_content("2022-06-01")
+    expect(page).to have_content("2022-04-01")
```
alicelieutier, line 35: This doesn't actually check the order of the content. You could use regular expressions here to check that "2022-08-01" is above "2022-06-01", for example with something like
```rb
expect(page).to match(/2022-08-01.*2022-06-01/)
```

However, I think an even better test would check the content of the mew message comes before the content of an older one, this way the test would still pass if you decide to change the formatting of the date, given that the user story is about posts order, not dates :)

PR #216, opened 2022-04-01
first version

Leigh-Rose:
# Your name

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,8 +1,20 @@
 require 'sinatra/base'
+require './lib/peeps'
 
 class Chitter < Sinatra::Base
-  get '/test' do
-    'Test page'
+
+  # get '/' do
+    # redirect '/chitter/peeps'
+  # end
+
+  get '/chitter/peeps' do
+    @peeps = Peeps.all
+    erb :'chitter/peeps'
+  end
+
+  post '/chitter/peeps' do
+    Peeps.create(message: params[:message])
+    redirect :'chitter/peeps'
   end
 
   run! if app_file == $0
--- a/lib/peeps.rb
+++ b/lib/peeps.rb
@@ -1,0 +1,31 @@
+class Peeps
+  attr_reader :id, :message
+
+  def initialize(id:, message:)
+    @id = id
+    @message = message
+  end
+
+  def self.all
+    if ENV['ENVIRONMENT'] == 'test'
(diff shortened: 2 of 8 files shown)


ghost:
Nice use of user prompt in ‘new peeps’ form ‘Please enter a new peep below:’

App has also been secured with the use of exec_params 

code is well laid out & easy to read - unnecessary files have been removed to make the files clear (such as the pre-given test page spec)

Naming clear for files such as view/add

app.rb
```diff
@@ -1,8 +1,20 @@
 require 'sinatra/base'
+require './lib/peeps'
 
 class Chitter < Sinatra::Base
-  get '/test' do
-    'Test page'
+
+  # get '/' do
```
tomfuller, line 6: I would probably try to avoid leaving commented out code in your code base, either you need that redirect or you don't

app.rb
```diff
@@ -1,8 +1,20 @@
...
+    # redirect '/chitter/peeps'
+  # end
+
+  get '/chitter/peeps' do
```
tomfuller, line 10: It looks like you tried to organise your routes, however the best way to do this is with a system we call REST. For example with your routes as they are now you would presumably have /chitter/ in front of everything making it unhelpful bloat, and peeps is a bit generic and doesn't give any indication of what that route is for. Perhaps something like get '/peeps' for the landing page and then post '/peeps/new' might be a better solution for your other route. This way we know that these routes are for peeps and that one of them is to create a new one.

lib/peeps.rb
```diff
@@ -0,0 +1,31 @@
+class Peeps
+  attr_reader :id, :message
+
+  def initialize(id:, message:)
+    @id = id
+    @message = message
+  end
+
+  def self.all
+    if ENV['ENVIRONMENT'] == 'test'
```
tomfuller, line 10: This if else is going to be needed in any function you create and it already adds quite a lot of bloat with just two at the moment, could you extract that into a separate function?

tomfuller:
Overall this is a nice solution to the problem, well done! I have left some comments where there are a couple things that could be improved, but I also wanted to call out that while having the UI tests is great it would also be good to have unit tests testing your individual functions. If you have any questions about any of the feedback let me know and Im happy to chat it through 😄

PR #220, opened 2022-04-04
Viktor' Entry

hamolicious:
# Your name
Viktor Patmalnieks

# User stories 

# README checklist

- [x] how to install,
- [x] how to run,


app.rb
```diff
@@ -1,9 +1,29 @@
...
+  # setup sinatra
+  set :root,  File.dirname(__dir__)
+  set :views, proc { File.join(__dir__, 'lib/views') }
+  set :public_folder, proc { File.join(__dir__, 'lib/assets') }
+
+  enable :sessions
+
   get '/test' do
     'Test page'
   end
```
EdwardAndress, line 16: This is great but there are some unconventional names and it's worth reflecting on that.  I'd expect the route for getting all Posts to be `GET '/posts'`. You can potentially use the same route for getting the filtered list, if you add a little bit of logic to the controller action.

I'd expect the method for getting all posts to be `Post.all`

lib/models/databaseconnection.rb
```diff
@@ -0,0 +1,30 @@
+require 'pg'
+
+class DatabaseConnection
+  attr_accessor :host, :name, :user
+
+  @@host = String('0.0.0.0')
```
EdwardAndress, line 6: Creating Strings like this is a little unconventional in Ruby

lib/models/databaseconnection.rb
```diff
@@ -0,0 +1,30 @@
...
+  def self.open
+    @@con = PG::Connection.new(host: @@host, dbname: @@name, port: @@port)
+  end
+
+  def self.close
+    @@con.close
+    @@con = nil
+  end
+
+  def self.query(q)
```
EdwardAndress, line 24: Why do you open and close the connection around each query?

lib/models/post.rb
```diff
@@ -0,0 +1,66 @@
+require_relative './databaseconnection'
+require_relative './user'
+
+class Post
+  attr_reader :id, :body, :created_on, :user_id
+
+  # User object methods
+  def initialize(hash)
+    @id = hash.fetch('id')
```
EdwardAndress, line 9: you can do `hash['id']` to fetch the id.  No need for the `fetch` method here

lib/models/post.rb
```diff
@@ -0,0 +1,66 @@
...
+    @body = hash.fetch('body')
+    @created_on = hash.fetch('created_on')
+    @user_id = hash.fetch('user_id')
+  end
+
+  def get_postee
+    User.get_by_id(@user_id)
+  end
+
+  def to_h
```
EdwardAndress, line 19: I'd advise against creating methods that are purely for your tests

lib/models/post.rb
```diff
@@ -0,0 +1,66 @@
...
+    for res in r.to_a
+      posts.push(Post.new(res))
+    end
+    posts
+  end
+
+  def self.search_bodies(keywords)
+    query = 'SELECT * FROM peeps WHERE '
+
+    for kw in keywords
```
EdwardAndress, line 46: This is very neat!

lib/models/post.rb
```diff
@@ -0,0 +1,66 @@
...
+      query << "body LIKE ('%#{kw}%') OR "
+    end
+    query.chomp!(' OR ')
+
+    r = DatabaseConnection.query(query)
```
EdwardAndress, line 51: I'd advise against the use of single letter variable names. They'll confuse frustrate your colleagues :)

PR #222, opened 2022-04-04
Chitter Challenge

jkgiwa:
# Your name

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,


peep.rb
```diff
@@ -0,0 +1,29 @@
+require 'pg'
+
```
EdwardAndress, line 2: detail: class names should always start with a capital letter

app.rb
```diff
@@ -1,9 +1,20 @@
...
   get '/test' do
     'Test page'
   end
 
+  get '/' do
+    'Welcome to Chitter'
+  end
+
+  get '/peeps' do
+    Chitter.add
```
EdwardAndress, line 14: This route should return a template that shows the user all of the peeps in the database. That means you'll need to use Peep.all at some point.

EdwardAndress:
Hi there! You've made a decent start to this challenge but there are a few lines that will, I think, throw errors. For example there is no `Chitter.add` method as far as I can see, so line 14 of `app.rb` will throw an error.

Speculating a little...

I think that you might be trying to build the whole application at once, which is always going to be hard, even for experienced developers. Everything will become much easier if you can get into the habit of building applications in very small steps, validating each one as you go. The process workshops which get run from time to time are an excellent place to practice this.

I hope this makes sense and is helpful. If you have any questions, please reach out to me on Slack.

PR #230, opened 2022-07-15
Kyle Tarbuck - Chitter Challenge

ghost:
# Your name

Kyle Michael Tarbuck

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,


ghost (commented):
Had added everything to an index file initially but think this will be better as a new view which can be posted to from the index submit form.

spec/features/peep_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'peep'
+
+describe '.all' do
+  it 'returns all peeps' do
+    connection = PG.connect(dbname: 'chitter_test')
```
Marlon-Buckley, line 5: This looks good, tested locally and method works with a 'chitter_test' db.

views/view_peeps.erb
```diff
@@ -0,0 +1,6 @@
+<h1>CHECK OUT THE LATEST PEEPS!</h1>
```
Marlon-Buckley, line 1: This is good and works as expected, you could now add a feature test in for this page.

app.rb
```diff
@@ -1,8 +1,14 @@
...
 
 class Chitter < Sinatra::Base
   get '/test' do
-    'Test page'
+    "Testing testing"
+  end
+
+  get '/view_peeps' do
+    @peeps = Peep.all
+    erb :view_peeps
```
Marlon-Buckley, line 11: I think convention says we should use hyphens in the url so that would also need to match with the name of the views erb file.

app.rb
```diff
@@ -1,8 +1,14 @@
 require 'sinatra/base'
+require './lib/peep'
 
 class Chitter < Sinatra::Base
   get '/test' do
-    'Test page'
+    "Testing testing"
+  end
```
siellsiell, line 7: It's always a good idea to remove test code from code you submit for code review (anywhere, not just at Makers)

views/view_peeps.erb
siellsiell: Agree, it would be good to see a feature test in addition to the nice unit test you've got!

siellsiell (commented):
Hey Kyle,

This is a good start at the challenge. You've shown that you know how to apply the MVC model and how to get a webapp to read from a database. Next step would be to try your hand at writing to the database too and doing a bit more feature testing. Was there any point where you ran into any blockers?

I left a few more comments below, let me know if you have any questions! You can answer here in the comments or DM me.

PR #231, opened 2022-07-15
Chitter solo challenge

olliefreeman94:
# Your name

Ollie Freeman

# User stories 

# README checklist

- [x] how to install,
- [x] how to run,


app.rb
```diff
@@ -1,9 +1,38 @@
...
+  get '/peeps/search' do
+    erb :'peeps/search'
+  end
+
+  get '/peeps/filtered' do
+    @peeps = Peep.filter(params[:keyword])
+    @view_all_button = 1
+    erb :'peeps/index'
+  end
+
```
Joemusarurwa, line 36: clean and concise. I can follow the workflow from how the methods are structured.

notes.txt
```diff
@@ -0,0 +1,13 @@
...
+- consolidate helper classes into one file?
+- merge or link '/' and 'peeps' routes?
+- pathing between routes
+
+Additional features
+- record time as well as date
+- message character limit validation
+- cf. Bookmark Manager extensions,
+    e.g. update (inc. logging?), delete, user authentication
+- equivilent of Twitter #
```
Joemusarurwa, line 13: good use of notes and to do to plan next steps

views/peeps/new.erb
```diff
@@ -0,0 +1,7 @@
+<h1>New peep</h1>
+
+<form action="/peeps" method="post">
+  <input type="text" name="message" placeholder="Message"/>
+  <input type="hidden" name="date" value="<%=Date.today%>">
```
Joemusarurwa, line 5: awesome way to get the time stamp with a hidden field

README.md
```diff
@@ -1,66 +1,42 @@
-## Chitter Challenge
```
siellsiell, line 1: Nice README, it's very clear.

app.rb
```diff
@@ -1,9 +1,38 @@
 require 'sinatra/base'
+require_relative './lib/peep'
 
 class Chitter < Sinatra::Base
   get '/test' do
     'Test page'
   end
```
siellsiell, line 7: It's always a good idea to remove test code from code you submit for code review (anywhere, not just at Makers)

app.rb
```diff
@@ -1,9 +1,38 @@
...
+
+  get '/peeps/search' do
+    erb :'peeps/search'
+  end
+
+  get '/peeps/filtered' do
+    @peeps = Peep.filter(params[:keyword])
+    @view_all_button = 1
+    erb :'peeps/index'
+  end
```
siellsiell, line 35: You've shown good use of RESTful routes 👍🏽

app.rb
```diff
@@ -1,9 +1,38 @@
...
+    erb :'peeps/new'
+  end
+
+  get '/peeps/search' do
+    erb :'peeps/search'
+  end
+
+  get '/peeps/filtered' do
+    @peeps = Peep.filter(params[:keyword])
+    @view_all_button = 1
```
siellsiell, line 33: I saw further down that you're using this as a kind of boolean flag, in which case it would be better to make your intention explicit here and set this to true rather than one.

db/migrations/02_add_date_to_peeps.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE peeps ADD COLUMN date DATE;
```
siellsiell, line 1: Separate migration file -- great. And DATE is an appropriate datatype to use. Another option would have been to use TIMESTAMP. That would give you flexibility in the future if it ever becomes necessary to show time information too.

lib/peep.rb
```diff
@@ -0,0 +1,50 @@
...
+    @message = message
+    @date = date
+  end
+
+  def self.all
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
```
siellsiell, line 18: You're using this same piece of code in more than one place in this file. How might you get rid of this duplication?

lib/peep.rb
```diff
@@ -0,0 +1,50 @@
...
+  def self.filter(keyword)
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
+    
+    result = connection.exec(
+      "SELECT * FROM peeps WHERE message ILIKE '%#{keyword}%' ORDER BY date DESC"
+      )
```
siellsiell, line 47: Nice use of ILIKE!

spec/features/creating_messages_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'timecop'
+
+feature 'adding a new message' do
+  scenario 'a user can add a message to Chitter' do
+    Timecop.freeze(Time.local(2022, 7, 15)) do
```
siellsiell, line 5: Cool! Timecop is useful. Another option that doesn't require a gem would have been to stub the return value of `Time.now`, e.g.

```
@date = Time.local(2022, 7, 15)
allow(Time).to receive(:now).and_return(@date)
```

spec/features/filtering_messages_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+    Peep.add(message: 'Banana peep', date: '2022-07-15')
+    Peep.add(message: 'Apple', date: '2022-07-10')
+    Peep.add(message: 'Another banana peep', date: '2022-07-05')
+    
+    visit('/peeps/search')
+    fill_in('keyword', with: 'banana')
+    click_button('Search')
+
+    expect(page).to have_content "Banana peep (2022-07-15)"
+    expect(page).to have_content "Another banana peep (2022-07-05)"
```
siellsiell, line 12: This is good but doesn't exclude the possibility that the other non-banana peeps are also on the page! How could you test for that?

spec/features/pathing_spec.rb
```diff
@@ -0,0 +1,43 @@
+feature 'pathing between routes' do
```
siellsiell, line 1: This is very thorough, nice!

spec/features/viewing_messages_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    expect(page).to have_content "This is a third peep!"
+  end
+
+  scenario 'a user can see messages sorted reverse chronologically' do
+    Peep.add(message: 'Old peep', date: '2020-01-01')
+    Peep.add(message: 'New peep', date: '2021-01-01')
+    
+    visit('/peeps')
+
+    expect(page).to have_content(/New peep.*Old peep/)
```
siellsiell, line 26: Nice way to test this.

spec/units/peep_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      Peep.add(message: 'This is also a peep!', date: '2022-07-10')
+      Peep.add(message: 'This is a third peep!', date: '2022-07-05')
+      
+      peeps = Peep.all
+
+      expect(peeps.length).to eq 3
+      expect(peeps.first).to be_a Peep
+      expect(peeps.first.id).to eq peep.id
+      expect(peeps.first.message).to eq("This is a peep!")
+      expect(peeps.first.date).to eq("2022-07-15")
```
siellsiell, line 17: Since the peeps array could have duplicates (due to a bug), it's also necessary to test that the other 2 peeps are the ones you expect.

views/peeps/new.erb
siellsiell, line 5: Using a hidden field works but it does mean that a malicious user could issue their own POST request and set a time of their choosing. The other pitfall is that the date will be generated at the time when this page is rendered rather than at the time when the user clicks submit, so at midnight boundaries it might end up being wrong.

For reasons like that it can be better to generate the date in your Model code just before you insert the data into the database.

siellsiell (commented):
Hey Ollie,

This is a great solution to the challenge, really well done. Your testing in particular is very thorough, which is nice to see.

I've left a few more detailed comments below for things you could do in the future to make your code even better.

If you have any questions, feel free to ask in the comments on here or in a DM.

Simo

PR #234, opened 2022-07-15
Joe M - Chitter 

Joemusarurwa:
# Joe Musarurwa

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,


README.md
```diff
@@ -34,18 +34,25 @@ You should see 1 passing test.
...
+
 ```
 As a Maker
 So that I can see what people are doing
 I want to see all the messages (peeps)
 in a browser
 ```
 
+
+![alt text](story1diagram.png "See all messages diagram")
```
olliefreeman94, line 46: I really like the diagrams, they're super helpful for visualising the data flow in your program

app.rb
```diff
@@ -1,9 +1,34 @@
...
+
+  get '/new' do
+    erb :'new'
+  end
+
+  post '/' do
+    message = params['message']
+    p params['message']
+    time = Time.now.httpdate
+    connection = PG.connect(dbname: 'chitter')
```
olliefreeman94, line 24: One next step here would be to refactor this code into your Peeps class, similar to this step from the Bookmark Manager challenge: https://github.com/makersacademy/course/blob/main/apprenticeships_bookmark_manager/walkthroughs/10.md#refactoring

lib/peeps.rb
```diff
@@ -0,0 +1,18 @@
...
+  def self.all
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
+
+    result = connection.exec('SELECT message, posted FROM peeps')
+    # binding.irb
+    result.map { |peep| peep }
```
olliefreeman94, line 15: This part of the Bookmark Manager walkthrough outlines wrapping the database data in an instance of the class: https://github.com/makersacademy/course/blob/main/apprenticeships_bookmark_manager/walkthroughs/11.md#wrapping-returned-data

views/index.erb
```diff
@@ -0,0 +1,7 @@
+<ul>
+  <% @peeps.each do |peep| peep['message'] %>
+    <li><%= peep %></li>
+  <% end %>
+</ul>
+
+<img src='/chitter.png' />
```
olliefreeman94, line 7: The logo is a really nice addition!

README.md
```diff
@@ -34,18 +34,25 @@ You should see 1 passing test.
```
olliefreeman94, line 34: I found this framework really helpful when writing my README: https://github.com/makersacademy/course/blob/main/pills/readmes.md

db/migrations/01_create_chitter_table.sql
```diff
@@ -1 +1 @@
-CREATE TABLE peeps(id SERIAL PRIMARY KEY, message VARCHAR(60));
+CREATE TABLE peeps(id SERIAL PRIMARY KEY, message VARCHAR(60), posted VARCHAR(60));
```
olliefreeman94, line 1: I'm not sure I 100% follow their reasoning, but Makers advise adding updates to the database table as a separate migration file: https://github.com/makersacademy/course/blob/main/apprenticeships_bookmark_manager/walkthroughs/11.md#updating-the-bookmarks-table

Also, depending on how you record the date / time, you could use a different SQL data type for the _posted_ column: https://www.w3schools.com/sql/sql_datatypes.asp

views/index.erb
```diff
@@ -0,0 +1,7 @@
+<ul>
+  <% @peeps.each do |peep| peep['message'] %>
+    <li><%= peep %></li>
```
olliefreeman94, line 3: This part of the Bookmark Manager walkthrough outlines updating the view to show different properties of the Peeps object: https://github.com/makersacademy/course/blob/main/apprenticeships_bookmark_manager/walkthroughs/11.md#showing-title-in-the-bookmarks-view

olliefreeman94 (commented):
I wasn't able to test the code locally, so haven't been able to look at everything outlined here: https://github.com/makersacademy/course/blob/main/pills/code_review_checklist.md

The code is really clear, easy to follow, and uses the MVC paradigm well.

README.md
siellsiell: I agree! It's always a good idea to diagram things out.

One thing I would note is that in these diagrams you've included both `app.rb` and `peeps.rb` under the Controller and under model you have the database itself.

That's not quite how the MVC model is meant to be understood. The database isn't really a part of the MVC model, the MVC model only covers the code you write:

- The Model part refers to the code you write that interacts with the database (peeps.rb) and that provides a way to represent the data once you've retrieved it from the database (one instance of the `Peep` class represents a single row of the DB)
- As your diagram already suggests, the Controller part refers to the code you write that sets up routes and is the link between your View and your Model. It's just that in your codebase, this means that the Controller is really only the code in the `app.rb` file (in more complex codebases it could cover more files, but there's no need to worry about that at this point in the course!)

app.rb
```diff
@@ -1,9 +1,34 @@
 require 'sinatra/base'
+require './lib/peeps'
+require 'time'
 
 class Chitter < Sinatra::Base
   get '/test' do
     'Test page'
   end
```
siellsiell, line 8: It's always a good idea to remove test code from code you submit for code review (anywhere, not just at Makers). It keeps everything nice and clean :)

app.rb
siellsiell: Yes, that would be a good next step. It would allow you to test this code in a unit test just like you tested the `Peep.all` method in `peeps_spec.rb`. 
It would also make the separation between Controller and Model that I talked about above clearer.

spec/features/new_peep_spec.rb
```diff
@@ -0,0 +1,11 @@
+# in spec/features/creating_bookmarks_spec.rb
+
+feature 'Post a new peep' do
+  scenario 'A user can post a new peep to chitter' do
+    visit('/new')
+    fill_in('message', with: 'hey peeps. just a peep to say I am posting a new peep. so peep this!')
+    click_button('post your peep')
+
+    expect(page).to have_content 'hey peeps. just a peep to say I am posting a new peep. so peep this!'
+  end
```
siellsiell, line 10: Good feature test!

spec/features/view_peeps_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+  scenario 'A user can see peeps' do
+    connection = PG.connect(dbname: 'chitter_test')
+
+    # Add the test data
+    connection.exec("INSERT INTO peeps VALUES(1, 'Just completed SQL Zoo. It was fun to do!');")
+    connection.exec("INSERT INTO peeps VALUES(2, 'Testing adding chitter messages');")
+    connection.exec("INSERT INTO peeps VALUES(3, 'testing chitter.');")
+    
+    visit('/')
+    expect(page).to have_content "Just completed SQL Zoo. It was fun to do!"
```
siellsiell, line 11: 💥

db/migrations/01_create_chitter_table.sql
siellsiell: Some good links there @olliefreeman94 

The reason for a separate migration file is that in a real system, you would want to be able to run the migration separately from the initial creation of the table, as a DB will typically throw an error if you try to create a table that doesn't exist. If all the migrations are in the same file, they will never run because the error will happen before we even get to them.

siellsiell, line 1: When we want to add a field to a new table, we typically want to use an `ALTER TABLE` statement rather than create a brand new table. Think about what would happen -- all of the existing peeps would no longer be visible on the site because the new table doesn't contain them. The link @olliefreeman94 sent above has a good example of that.

spec/peeps_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+      connection.exec("INSERT INTO peeps VALUES(1, 'Just completed SQL Zoo. It was fun to do!');")
+      connection.exec("INSERT INTO peeps VALUES(2, 'Testing adding chitter messages');")
+      connection.exec("INSERT INTO peeps VALUES(3, 'testing chitter.');")
+  
+
+      peeps = Peeps.all
+
+      expect(peeps).to include "Just completed SQL Zoo. It was fun to do!"
+      expect(peeps).to include "Testing adding chitter messages"
+      expect(peeps).to include "testing chitter."
```
siellsiell, line 20: Nice and thorough!
Even more thorough: checking that `peeps` has the right length -- what if some duplicate messages had crept in somehow?

siellsiell (commented):
Hey Joe,

Well done on completing the first two user stories. The diagrams were a great addition. I left a few more detailed comments below. I would suggest using this week where you'e building another database app to revisit some of the ideas of Database week, namely:
- the difference between Model and Controller
- how we can update existing tables in a database using migrations

Using Makersbnb as a way to revisit these topics would be a great use of this week. Makersbnb is intentionally an app quite similar to the one's you've built so far so that you can consolidate your learning from the previous weeks. The only new thing is the teamwork aspect.

If you have questions about anything I commented on, feel free to ask me here in the comments or on Slack!

Simo

PR #235, opened 2022-07-15
Miranda Frudd - chitter challenge (unfinished)

MirandaFrudd:
# Your name

Miranda Frudd

# User stories 

- [X] User story 1: "I want to see all the messages (peeps) in a browser"

# README checklist

- [X] how to install,
- [X] how to run,
- [X] and how to test your code?

-> yes but I've mainly used what was already given to us


spec/features/messages_page_spec.rb
```diff
@@ -0,0 +1,9 @@
+feature '/' do
+  scenario 'to submit a form' do
+    visit '/'
+    fill_in 'Message', with: 'First Message'
+    click_on 'Submit'
+    visit '/messages'
+    expect(page).to have_content "First Message"
```
siellsiell, line 7: This is a good example of a feature test, well done!

spec/message_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    end
+  end
+end
+
+# RSpec.describe '.create' do
+#   context "so that I can create a new message" do
+#     it "#create adds a new message" do
+#       message = Message.create(message: 'Another message')
+#       expect(message.message).to eq "Another message"
```
siellsiell: You were really close! One way to check that `create` has really added the message to the database would be to call `Message.all` after calling `create` and check that your message is now part of the result set.

app.rb
```diff
@@ -1,8 +1,22 @@
...
-    'Test page'
+
+  enable :sessions
+
+  get '/' do
+    erb :index
+  end
+
+  post '/messages' do
+    # Message.create(message: params[:message])
```
siellsiell: It looks like you were thinking of creating the message in your POST route above (the commented out bit), which is good. Where would the right place be to retrieve it the message again so that you can display it to the user? You already have all the methods you need to achieve that in your `Message` class!

lib/message.rb
```diff
@@ -0,0 +1,33 @@
...
+    result.map { |peep| peep['message'] }
+    # return @messages
+  end
+
+  def self.create(message:)
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
```
siellsiell, line 19: You're using this same piece of code in more than one place in this file. How might you get rid of this duplication?

siellsiell (commented):
Hey Miranda,

I know you felt some confusion about how to connect an app to a database, but you were actually really close to connecting it all up -- all the ingredients are there. I left a few comments with some pointers. 

Let me know if you have any questions! You can reply to my comments on this PR or DM me.

Simo

spec/message_spec.rb
MirandaFrudd: Thank you! I think I understand how the class variables can be used to query the database now. I think I was trying to add class objects to the database but that's not right - I needed to gather params from a form and add them to a row of the database table using an SQL query within my class

app.rb
MirandaFrudd: I should have used Message.all to display the messages from the db table in the get route

siellsiell: Exactly! Nice one.

spec/message_spec.rb
siellsiell: Correct!

PR #236, opened 2022-07-18
15/07/22

erb16:
# Your name
Emily Bayliss

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,8 +1,35 @@
 require 'sinatra/base'
+require './lib/chitter'
 
 class Chitter < Sinatra::Base
   get '/test' do
     'Test page'
+  end
+
+  get '/' do
+    erb :index
+  end
+
+  get '/peeps' do
+    @peeps = Peep.all
+    # @time = Peep.time
+    erb :peeps
+  end
+
+  post '/peeps' do
+    Peep.create(peep: params[:peep], timestamp: params[:timestamp])
+
+    # post '/bookmarks' do
+    #   Bookmark.create(url: params[:url], title: params[:title])
+    #   redirect '/bookmarks'
+    # end
+    # @time = Peep.time
+    @peeps = Peep.all
+    erb :peeps
+  end
+
+  get '/peeps/add' do
+    erb :"/peeps/add"
   end
 
   run! if app_file == $0
--- a/db/migrations/02_add_timestamp_to_chitter.sql
+++ b/db/migrations/02_add_timestamp_to_chitter.sql
@@ -1,0 +1,2 @@
+-- To both chitter and chitter_test databases:
+ALTER TABLE peeps ADD COLUMN timestamp VARCHAR(60);
--- a/lib/chitter.rb
+++ b/lib/chitter.rb
@@ -1,0 +1,35 @@
+require 'pg'
+
+class Peep
+  def self.all
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
+
+    result = connection.exec("SELECT * FROM peeps")
+    result.map { |peep| peep['message'] }
+  end
+
+  def self.create(peep:, timestamp:)
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
+    
+    connection.exec("INSERT INTO peeps (message, timestamp) VALUES('#{peep}', #{timestamp})")
+    # connection.exec("INSERT INTO bookmarks (title, url) VALUES('#{title}', '#{url}') RETURNING id, url, title")
+  end
+
+  # def self.time
+  #   if ENV['ENVIRONMENT'] == 'test'
+  #     connection = PG.connect(dbname: 'chitter_test')
+  #   else
+  #     connection = PG.connect(dbname: 'chitter')
+  #   end
+
+  #   connection.exec("INSERT INTO peeps (timestamp) VALUES('#{Time.now}')")
+  # end
+end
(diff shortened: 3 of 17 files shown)


lib/chitter.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def self.create(peep:, timestamp:)
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
+    
+    connection.exec("INSERT INTO peeps (message, timestamp) VALUES('#{peep}', #{timestamp})")
```
Amy-O, line 22: nice and easy to read

spec/features/adding_peeps_spec.rb
```diff
@@ -0,0 +1,8 @@
+feature 'Viewing peeps' do
+  scenario 'viewing related timestamp' do
```
Amy-O, line 2: It seems like this test is just for viewing the peeps rather than timestamps, as the expected outcome is just a lil peep peep. I'm guessing you were halfway through adapting an existing test?

spec/features/viewing_peeps_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    # connection = PG.connect(dbname: 'chitter_test')
+
+    # # Add the test data
+    # connection.exec("INSERT INTO peeps VALUES(1, 'Peep one');")
+    # connection.exec("INSERT INTO peeps VALUES(2, 'Peep two');")
+    # connection.exec("INSERT INTO peeps VALUES(3, 'Peep three');")
+
+    # Bookmark.create(website: "http://www.makersacademy.com")
+    # Bookmark.create(website: "http://www.destroyallsoftware.com")
+    # Bookmark.create(website: "http://www.google.com")
```
Amy-O, line 15: I didn't think to add the bookmark stuff in comments for easier comparison! smart

app.rb
```diff
@@ -1,9 +1,36 @@
...
     'Test page'
   end
 
+  get '/' do
+    erb :index
+  end
+
+  get '/peeps' do
+    @peeps = Peep.all
+    # @time = Peep.time
```
siellsiell, line 15: It's always a good idea to remove commented out code from code you submit for code review (anywhere, not just at Makers)

db/migrations/02_add_timestamp_to_chitter.sql
```diff
@@ -0,0 +1,2 @@
+-- To both chitter and chitter_test databases:
+ALTER TABLE peeps ADD COLUMN timestamp VARCHAR(60);
```
siellsiell, line 2: Nice use of the timestamp column type!

lib/chitter.rb
```diff
@@ -0,0 +1,35 @@
...
+    result = connection.exec("SELECT * FROM peeps")
+    result.map { |peep| peep['message'] }
+  end
+
+  def self.create(peep:, timestamp:)
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
```
siellsiell, line 20: You're using this same piece of code in more than one place in this file. How might you get rid of this duplication?

spec/features/adding_peeps_spec.rb
siellsiell: Nice observation @Amy-O ! I agree, right now this test is not testing what it says it does.

spec/features/timestamp_rspec.rb
```diff
@@ -0,0 +1,8 @@
+feature 'Adding a peep' do
+  scenario 'adding a peep' do
+    visit('/peeps/add')
+    fill_in('peep', with: 'Peep peep!')
+    click_button('Post')
+    expect(page).to have_content ""
```
siellsiell, line 6: Shouldn't we be expecting the page to have "Peep peep!" as content?

spec/spec_helper.rb
```diff
@@ -41,6 +42,11 @@
 RSpec.configure do |config|
   config.before(:each) do
     setup_test_database
+
+#     new_time = Time.local(2008, 9, 1, 12, 0, 0)
+# Timecop.freeze(new_time)
```
siellsiell, line 47: I see you were trying to use Timecop to freeze the time during tests. Maybe it was tricky getting it to work? An alternative that doesn't require an additional gem is to use mocking, e.g.:

```
@now = Time.now
allow(Time).to receive(:now).and_return(@now)
```

app.rb
```diff
@@ -1,9 +1,36 @@
...
+    erb :peeps
+  end
+
+  post '/peeps' do
+    Peep.create(peep: params[:peep], timestamp: params[:timestamp])
```
siellsiell, line 20: Passing the time in here is one way to do it but it's vulnerable to bugs. For example:
- the user may click "submit" some time after the page was generated from your `erb` view, so the time of the peep will be inaccurate
- the user can influence the time by sending their own POST requests and putting in a bogus time (e.g. to get their peep to show first even though it's old)

For such reasons, it's best to generate timestamps in your Model when you're about to add the peep to the database so that you're sure it will be accurate. I see that you were thinking along those lines in your commented out code in the `chitter.rb`. You were on the right track!

To control the time in tests, you can stub the return value of `Time.now` (see another comment further down).

siellsiell (commented):
Hey Emily,

This is a good attempt at this challenge, well done! You've got good feature tests. Next time, make sure you add unit tests as well so that you can test your Peep class in isolation. I've left some more detailed comments below.

Simo

views/peeps.erb
```diff
@@ -0,0 +1,5 @@
+<ul>
+  <% @peeps.each do |peep| %>
+    <li><%=peep%></li>
```
siellsiell, line 3: All the rest of your code is there to fulfil user story 3 (showing the timestamp) -- did you run out of time or did you get blocked?

PR #239, opened 2022-07-18
Challenge submission

Marlon-Buckley:
# Your name

Marlon Buckley

# User stories 

- [x] User story 5: "I want to filter on a specific keyword" (This returns posts correctly however doesn't equate for when the keyword provided doesn't exist in any posts!)

# README checklist

- [x] how to install,
- [x] how to run,


app.rb
```diff
@@ -1,9 +1,42 @@
+require 'rack/handler'
 require 'sinatra/base'
+require './lib/peeps'
 
 class Chitter < Sinatra::Base
   get '/test' do
     'Test page'
   end
```
siellsiell, line 8: Before submitting code for review, it's good practice to remove any test code, commented out code and debugging stuff.

lib/peeps.rb
```diff
@@ -0,0 +1,68 @@
...
+      Peeps.new(id: peep['id'], peep: peep['message'], time: peep['time'])
+    end
+  end
+
+  def self.create(peep:, time:)
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
```
siellsiell, line 31: You're using this same piece of code in more than one place in this file. How might you get rid of this duplication?

lib/peeps.rb
```diff
@@ -0,0 +1,68 @@
...
+
+    result = connection.exec("SELECT message FROM peeps WHERE message LIKE '%#{keyword}%';")
+    p result.class
+    
+    if result == [] #this isn't working currently.
```
siellsiell, line 60: I remember you asked this in the kickoff and got a great answer about using `to_a` to convert it to an array. It doesn't look like there is any more elegant way of doing it!

app.rb
```diff
@@ -1,9 +1,42 @@
...
+    # p @posts
+    erb(:'posts/index')
+  end
+
+  get '/posts/new' do
+    erb :'posts/new'
+  end
+
+  post '/posts' do
+    Peeps.create(peep: params['peep'], time: params['time'])
```
siellsiell, line 21: Passing the time in here is one way to do it but it's vulnerable to bugs. For example:
- the user may click "submit" some time after the page was generated from your `erb` view, so the time of the peep will be inaccurate
- the user can influence the time by sending their own POST requests and putting in a bogus time (e.g. to get their peep to show first even though it's old)

For such reasons, it's best to generate timestamps in your Model/Controller code so that you're sure it will be accurate.
To control the time in tests, you can mock the return value of `Time.now`, for example:

```
@now = Time.now
allow(Time).to receive(:now).and_return(@now)
```

spec/peeps_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  describe '.search_by_keyword' do
+    it "Returns posts that contain the keyword provided" do
+      peep = Peeps.create(peep: 'This is the first sort by time test peep!', time: Time.new)
+      Peeps.create(peep: 'This is the second sort by time test peep!', time: Time.new)
+      Peeps.create(peep: 'Some other rubbish!', time: Time.new)
+      Peeps.create(peep: 'Yet more rubbish!', time: Time.new)
+      Peeps.create(peep: 'And one more for good measure!', time: Time.new)
+      peeps = Peeps.search_by_keyword("rubbish")
+      expect(peeps.length).to eq 2
```
siellsiell, line 51: If there are two peeps in the list then you should check that both are what you expect or you might miss a bug where peeps get duplicated, for example.

spec/peeps_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+require 'database_helpers'
+
+describe Peeps do
+  describe '.all' do
+    it 'returns all chitter posts' do
+      connection = PG.connect(dbname: 'chitter_test')
+      time_when_tested = "2022-07-15 12:00:15"
+      peep = Peeps.create(peep: "Hello this is my unit test for .all", time: time_when_tested)
+      Peeps.create(peep: "Hello this another peep from my unit test", time: time_when_tested)
+      peeps = Peeps.all
```
siellsiell, line 11: You're missing a line that asserts something about the value of `peeps`. Currently this test doesn't test (assert) anything.

spec/features/search_by_keyword_spec.rb
```diff
@@ -0,0 +1,8 @@
+feature 'Returns posts containing keyword provided' do
+  scenario 'Enter keyword in form' do
+    visit('/posts')
+    fill_in('keyword', with: 'Svit')
+    click_button ('Search')
+    expect(page).to have_content 'Posts containing keyword:'
+  end
```
siellsiell, line 7: For completeness if would be good to add a test that asserts what the page should look like what happens when a keyword does match

spec/features/sort_by_time_spec.rb
```diff
@@ -0,0 +1,6 @@
+feature 'Sorting posts by date in descending order' do
+  scenario 'visiting the sort-by page' do
+    visit('posts/sort-by')
+    expect(page).to have_content "Posts sorted by date/time:"
```
siellsiell, line 4: Same here, this test doesn't actually tests that posts are really sorted.

siellsiell (commented):
Hey Marlon,

Well done on completing all of the user stories! Aside from one issue with the timestamps , you've got a clean separation between  Model, View and Controller and both feature tests and unit tests, which is great. One thing you can focus on for the future is to be mindful of your tests. Make sure that all of your tests are meaningful and test what they say they test.

 I've left some more detailed comments below.

Simo

db/migrations/02_add_time_to_peeps_table.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE peeps ADD COLUMN time VARCHAR(60);
```
siellsiell, line 1: A better fit would be to use a column of type TIMESTAMP. This has the benefit of the database enforcing data integrity: it will throw an error if anything other than a timestamp is added to that column.

